PR #1, opened 2016-05-14
fix some stuff

doy:


jvns:
i think this stuff is all in now

PR #5, opened 2016-09-03
[WIP] Add Mac OS support

dirk:
Things to do:

- [x] Set up platform-specific modules with their own implementations of the `CopyAddress` trait in the `lib` module.
- [x] Actually verify Mac OS implementation of `CopyAddress` works.
- [x] Add platform-specific implementations of functions in `dwarf` module.
- [x] Verify those platform-specific implementations in `dwarf` work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ gimli={git="https://github.com/fitzgen/gimli.git"}
 elf = "0.0.9"
 log = "0.3.6"
 env_logger = "0.3.4"
+mach = "0.0.5"
 
 [dev-dependencies]
 flate2 = "0.2.14"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,8 @@ elf = "0.0.9"
 log = "0.3.6"
 env_logger = "0.3.4"
 mach = "0.0.5"
+libproc = "0.1.0"
+object = "0.1.0"
 
 [dev-dependencies]
 flate2 = "0.2.14"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,8 @@ env_logger = "0.3.4"
 mach = "0.0.5"
 libproc = "0.1.0"
 object = "0.1.0"
+libarchive = "0.1.1"
+libarchive3-sys = "0.1.2"
 
 [dev-dependencies]
 flate2 = "0.2.14"
--- a/src/dwarf.rs
+++ b/src/dwarf.rs
@@ -338,7 +338,7 @@ mod obj {
     use self::object::Object;
     use std::fs;
     use std::io::Read;
-    use std::path::Path;
+    use std::path::{Path, PathBuf};
 
     use super::{Entry, get_all_entries};
 
@@ -353,8 +353,28 @@ mod obj {
         buf
     }
 
+    /// The plain Ruby binary location on Mac OS is normally a slim Mach-O
+    /// executable binary. However there's often a `libruby-static.a` nearby
+    /// in the file system that *does* have DWARF info in it.
+    fn guess_ruby_path(pid: usize) -> PathBuf {
+        let exec_path = libproc::libproc::proc_pid::pidpath(pid as i32).expect("Could not look up path for PID");
+        let exec_path = PathBuf::from(&exec_path);
+
+        let static_path = exec_path.join("../../lib/libruby-static.a").canonicalize()
+            .expect("Could not guess libruby-static.a path");
+
+        println!("{:?}", static_path);
+
+        if static_path.exists() {
+            static_path
+        } else {
+            exec_path
+        }
+    }
+
     pub fn get_dwarf_entries(pid: usize) -> Vec<Entry> {
-        let path = libproc::libproc::proc_pid::pidpath(pid as i32).expect("Could not look up path for PID");
+        let path = guess_ruby_path(pid);
+
         let file = open(path);
         let file = object::File::parse(&file);
 
--- a/src/dwarf.rs
+++ b/src/dwarf.rs
@@ -410,7 +410,6 @@ mod obj {
             if !name.ends_with(".o") {
                 continue
             }
-            println!("Reading entries from '{}'", name);
 
             let file_entries = get_entries_from_file(buf);
             entries.extend(file_entries);
@@ -445,7 +444,7 @@ mod obj {
         let path = guess_ruby_path(pid);
 
         if path.extension().unwrap() == "a" {
-            println!("Using archive '{}'", path.to_str().unwrap());
+            debug!("Using archive '{}'", path.to_str().unwrap());
             get_archive_entries(path)
         } else {
             let file = open(path);
--- a/src/dwarf.rs
+++ b/src/dwarf.rs
@@ -7,7 +7,7 @@ use byteorder::{NativeEndian, ReadBytesExt};
 use std::collections::HashMap;
 use std::io::Cursor;
 
-pub use self::obj::get_dwarf_entries;
+pub use self::obj::{get_executable_path, get_dwarf_entries};
 
 type HashMapFnv<K, V> = HashMap<K, V, BuildHasherDefault<FnvHasher>>;
 
@@ -284,6 +284,10 @@ mod obj {
     /// The parsed object file type.
     type File = elf::File;
 
+    pub fn get_executable_path(pid: usize) -> Result<PathBuf, String> {
+        Ok(PathBuf::from(format!("/proc/{}/exe", pid)))
+    }
+
     pub fn get_dwarf_entries(pid: usize) -> Vec<Entry> {
         let file_path = format!("/proc/{}/exe", pid);
         let file = open(&file_path);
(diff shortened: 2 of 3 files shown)


kamalmarhubi:
> Allow passing in arbitrary path to other Mach-O containing actual DWARF info (because Mac can be really annoying and do things like have the DWARF info be in a separate file from the actual executable).

This is also the case with linux! Right now, most distros' ruby binary *won't* work with this. We've been testing with binaries built ourselves (using rbenv) that include the debug info. So there's definitely a need for this there too.

kamalmarhubi:
This is really really cool! I know zero about low level OS X stuff. It's great that `pidpath` and `vm_read` exist.

src/dwarf.rs
```diff
@@ -327,6 +329,70 @@ mod obj {
     }
 }
 
+#[cfg(target_os="macos")]
+mod obj {
+    extern crate gimli;
+    extern crate libproc;
+    extern crate object;
```
kamalmarhubi: Seems `object` is something the gimli folks wrote to make cross-platform stuff easier. It should also work for elf files on linux. If you just `#[cfg]` the method of getting the path, then I think the other functions you've written here should work on both platforms. That's a cool cleanup we can do later. There's also an opportunity to replace reading with memory mapping, I think. Nice :-)

dirk:
> It's great that pidpath and vm_read exist.

They do exist, but there may be some issues with permissions/code-signing. However, I'm not going to worry about that problem until I actually get to it. (Never let worry get in the way of some recreational coding.)

> Right now, most distros' ruby binary _won't_ work with this. We've been testing with binaries built ourselves (using rbenv) that include the debug info. So there's definitely a need for this there too.

So what I ended up doing, in 39d6b83, is looking for the `lib/libruby-static.a` file in the same tree as wherever the Ruby executable is located. Then I used the [libarchive](https://github.com/chef/libarchive-rust) package to scan through that static archive and look for object files. The object files in there seem to be the ones that get built when you compile Ruby (`ruby.o`, `string.o`, `vm.o`, etc. ) and they _do_ have DWARF info in them! 🎉

dirk:
So this now properly reads the `ruby_current_thread` pointer and dereferences it. However I think I'm running into some issues with my reads going over page boundaries and therefore being insufficient for the structs it's trying to dereference. Will probably need to rewrite the `copy_address` implementation to read multiple pages if necessary.

```
$ lldb
Executable module set to "/usr/local/var/rbenv/versions/2.3.0/bin/ruby".
Architecture set to: x86_64-apple-macosx.
(lldb) attach $PID
(lldb) expr ruby_current_thread
(rb_thread_t *) $0 = 0x00007f98e8d00000

$ sudo RUST_BACKTRACE=1 RUST_LOG=debug ./target/debug/ruby-stacktrace top $PID
DEBUG:ruby_stacktrace::dwarf::obj: Using archive '/usr/local/var/rbenv/versions/2.3.0/lib/libruby-static.a'
DEBUG:ruby_stacktrace: get_nm_address: 100249b18
DEBUG:ruby_stacktrace: get_maps_address: 104e3b000
DEBUG:ruby_stacktrace: get_ruby_current_thread_address: 105084b18
DEBUG:ruby_stacktrace: copy_address_raw: addr: 105084b18
DEBUG:ruby_stacktrace: copy_address_raw: addr: 7f98e8d00000
thread 'main' panicked at 'index 1248 out of range for slice of length 1136', ../src/libcore/slice.rs:549
stack backtrace:
...
  11:        0x1074022bc - _<collections..vec..Vec<T> as core..ops..Index<core..ops..Range<usize>>>::index::hbb677cf67581aa0e
  12:        0x1073ffd24 - ruby_stacktrace::map_bytes_to_struct::ha7333de96673eadb
  13:        0x1073c4e49 - ruby_stacktrace::copy_address_dynamic::he62ab08be2b84b6c
  14:        0x1073c2455 - ruby_stacktrace::get_cfps::hc41d2a40917c9e42
  15:        0x1073c1b4a - ruby_stacktrace::get_stack_trace::h678e77c623a4ca82
  16:        0x1073bb063 - ruby_stacktrace::main::hbfdbe29ce98a3b0a
```

Feel like this is a good enough stopping place for a first attempt at getting it working on Mac OS. Will do some refactoring and clean-up in a bit.

@kamalmarhubi, @jvns: Anything y'all particularly would want in this PR for mergability?

jvns:
This is awesome!

Right now I can't build this branch on my Linux machine -- I get this error from `cargo build`

```
Build failed, waiting for other jobs to finish...
failed to run custom build command for `libarchive3-sys v0.1.2`
Process didn't exit successfully: `/home/bork/work/ruby-stacktrace/target/debug/build/libarchive3-sys-eb2e687e6dc47562/build-script-build` (exit code: 101)
--- stderr
thread '<main>' panicked at 'Unable to locate libarchive, err="`\"pkg-config\" \"--libs\" \"--cflags\" \"libarchive\"` did not exit successfully: exit code: 1\n--- stdout\nPackage libarchive was not found in the pkg-config search path.\nPerhaps you should add the directory containing `libarchive.pc\'\nto the PKG_CONFIG_PATH environment variable\nNo package \'libarchive\' found\n"', /home/bork/.cargo/registry/src/github.com-88ac128001ac3a9a/libarchive3-sys-0.1.2/build.rs:32
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

as long as it builds and doesn't break the Linux version, I'm happy to merge anything :)

kamalmarhubi:
> Right now I can't build this branch on my Linux machine...

For debian / ubuntu, install libarchive-dev. There are a bunch of other build errors though, once you get past that.

@dirk can you attempt a build with `cargo build --target=x86_64-unknown-linux-gnu`? You may have to `rustup target add x86_64-unknown-linux-gnu` first. It should get you to see all the missing import errors after the module shuffling, even though it may not be able to link the final output on OS X.

Cargo.toml
```diff
@@ -18,6 +18,11 @@ gimli={git="https://github.com/fitzgen/gimli.git"}
 elf = "0.0.9"
 log = "0.3.6"
 env_logger = "0.3.4"
+mach = "0.0.5"
+libproc = "0.1.0"
+object = "0.1.0"
+libarchive = "0.1.1"
+libarchive3-sys = "0.1.2"
```
kamalmarhubi: could you move all the deps you added to a `[target.'cfg(target_os="macos")'.dependencies]` section? It would stop requiring libarchive on Linux where it's not needed, or at least not yet needed.
`

src/lib.rs
```diff
@@ -122,15 +125,32 @@ mod platform {
             let task = task_for_pid(self.pid);
             if task.is_err() { return task.map(|_| ()) }
 
-            unsafe {
-                let mut read: mach_msg_type_number_t = 0;
+            let page_addr      = (addr as i64 & (-4096)) as mach_vm_address_t;
+	        let last_page_addr = ((addr as i64 + buf.len() as i64 + 4095) & (-4096)) as mach_vm_address_t;
+            let page_size      = last_page_addr as usize - page_addr as usize;
```
kamalmarhubi: could you add a comment explaining what's going on here? looks like the interface for `vm_read` is quite different from `process_vm_readv`. you have to read full pages?

kamalmarhubi: found some info:
- https://opensource.apple.com/source/xnu/xnu-517.3.7/osfmk/man/vm_read.html
- https://www.gnu.org/software/hurd/gnumach-doc/Data-Transfer.html#Data-Transfer

Looks like it must be a round number of pages. It also sounds like you should call `vm_deallocate` on the pointer you get back, or else leak the memory.

Kind of crappy that the interface requires copying, but what to do. There *is* a `vm_overwrite` that allows copying to a specified buffer, but since it has to be full pages we can't copy to the passed-in buffer anyway.

Looks

kamalmarhubi:
@dirk 
> Feel like this is a good enough stopping place for a first attempt at getting it working on Mac OS. Will do some refactoring and clean-up in a bit.

> @kamalmarhubi, @jvns: Anything y'all particularly would want in this PR for mergability?

Could you quickly summarise what the state is at the stopping place? Unclear how close to working it is!

Cargo.toml
dirk: Had a mental note to move that over behind a target platform flag... 😄

src/lib.rs
dirk: Yeah I have a bunch of those [opensource.apple.com/source/xnu](https://opensource.apple.com/source/xnu) links bookmarked. (So fortunate for those FOSS licenses ensuring these header & source files are public.) Will put together a comment documenting how all these pieces (I think) fit together. Will also make sure that `vm_deallocate` gets called once we copy the buffer over into the program's memory.

dirk:
> can you attempt a build with `cargo build --target=x86_64-unknown-linux-gnu`? You may have to `rustup target add x86_64-unknown-linux-gnu` first. It should get you to see all the missing import errors after the module shuffling, even though it may not be able to link the final output on OS X.

I'll give that a shot. May also try firing up an Ubuntu'ish VM locally to get the most thorough repro.

> Could you quickly summarise what the state is at the stopping place? Unclear how close to working it is!

Can do!

kamalmarhubi:
Re linux build, I'm also happy to fix it up, since it's almost certainly pretty minor. Let me know if that's easier!

jvns:
going to quickly close & reopen to see if I can get Travis to build this PR

kamalmarhubi:
Once more after adding the libarchive dep in travis

kamalmarhubi:
oh I made it require a rebase blarg

kamalmarhubi:
@dirk, merged this as #7 after rebase. Thanks!

PR #7, opened 2016-09-07
Add initial OS X support

kamalmarhubi:
This was originally #5 by @dirk. I rebased and made it build on Linux.


dirk:
@kamalmarhubi: Thanks so much for finishing this up! Been too busy with work the past couple days to get back to it. 😞

PR #10, opened 2016-09-13
Read across aligned pages on macOS and add some documentation

dirk:
The bad news is that macOS only lets you read a page at a time with `vm_read`. However, the good news is that—after reading some of the JDK's debugger source for that platform (see [here](https://github.com/openjdk-mirror/jdk7u-hotspot/blob/50bdefc/agent/src/os/bsd/MacosxDebuggerLocal.m#L131-L201))—I put together a `read_aligned` function to handle the messy work of reading multiple pages and smushing together the results.

The result of that is that `copy_address` is now concerned with just the addresses (and slicing into the result buffer), and `read_aligned` is concerned with just the reading. 🎉 

This still doesn't get it truly _working_ on macOS, but it's a bit more progress (and served as a fun little exercise for the evening).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -106,14 +106,14 @@ mod platform {
     #[allow(non_camel_case_types)] type vm_size_t = mach_vm_size_t;
 
     extern "C" {
-        /// Xnu system call to read the memory of a process.
+        /// XNU system call to read the memory of a process.
         ///
         /// Source of the `vm_read` function is at:
         ///   https://github.com/opensource-apple/xnu/blob/27ffc00/osfmk/vm/vm_user.c#L533-L549
         fn vm_read(target_task: vm_map_t, address: vm_address_t, size: vm_size_t, data: &*mut u8, data_size: *mut mach_msg_type_number_t) -> kern_return_t;
     }
 
-    /// Xnu/BSD function to gets the task port for the process identifier by
+    /// XNU/BSD function to gets the task port for the process identified by
     /// its PID.
     ///
     /// Source is at:
@@ -184,7 +184,7 @@ mod platform {
     }
 
     impl CopyAddress for Process {
-        /// Uses the Xnu `vm_read` function to copy a region of the process'
+        /// Uses the XNU `vm_read` function to copy a region of the process'
         /// memory, specified by `addr`, into `buf`.
         fn copy_address(&self, addr: usize, buf: &mut [u8]) -> io::Result<()> {
             let task = task_for_pid(self.pid);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -145,6 +145,9 @@ mod platform {
         ((x + VM_PAGE_MASK) & !VM_PAGE_MASK)
     }
 
+    /// Read multiple pages from a task/process's memory. Address arguments
+    /// *must* be page-aligned or else bad things will happen. For
+    /// page-alignment try `mach_vm_trunc_page` and `mach_vm_round_page`.
     fn read_aligned(task: mach_port_name_t, start_addr: mach_vm_address_t, end_addr: mach_vm_address_t) -> io::Result<Vec<u8>> {
         debug!("read_aligned: start_addr: {:x}", start_addr);
         debug!("read_aligned: end_addr: {:x}", end_addr);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,7 +93,7 @@ mod platform {
 
     use self::mach::kern_return::{kern_return_t, KERN_SUCCESS};
     use self::mach::port::{mach_port_t, mach_port_name_t, MACH_PORT_NULL};
-    use self::mach::vm_types::{mach_vm_address_t, mach_vm_size_t};
+    use self::mach::vm_types::{mach_vm_address_t, mach_vm_offset_t, mach_vm_size_t};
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -122,38 +131,81 @@ mod platform {
         Ok(task)
     }
 
-    impl CopyAddress for Process {
-        fn copy_address(&self, addr: usize, buf: &mut [u8]) -> io::Result<()> {
-            let task = task_for_pid(self.pid);
-            if task.is_err() { return task.map(|_| ()) }
+    const VM_PAGE_SIZE: vm_size_t = 4096;
+    const VM_PAGE_MASK: vm_size_t = 4095;
```
kamalmarhubi: maybe `!4095` to make masking not require the negation?

src/lib.rs
```diff
@@ -122,38 +131,81 @@ mod platform {
...
+
+        for i in 0..num_pages {
+            let page_addr = start_addr + (i * VM_PAGE_SIZE);
 
             let read_ptr: *mut u8 = ptr::null_mut();
             let mut read_len: mach_msg_type_number_t = 0;
 
             let result = unsafe {
-                vm_read(task.unwrap(), page_addr as u64, page_size as vm_size_t, &read_ptr, &mut read_len)
+                vm_read(task, page_addr as u64, VM_PAGE_SIZE as vm_size_t, &read_ptr, &mut read_len)
```
kamalmarhubi: cast of `VM_PAGE_SIZE` shouldn't be necessary

src/lib.rs
```diff
@@ -122,38 +131,81 @@ mod platform {
...
+            let read_buf = read_aligned(task.unwrap(), start_addr, end_addr);
+
+            match read_buf {
+                Err(err) => Err(err),
+                Ok(read_buf) => {
+                    // Get the offset into the page-aligned buffer
+                    let start = addr - start_addr as usize;
+                    let end = start + buf.len();
+
+                    buf.copy_from_slice(&read_buf[start..end]);
```
kamalmarhubi: sad this requires copying :/

src/lib.rs
```diff
@@ -122,38 +131,81 @@ mod platform {
...
+            if read_len != VM_PAGE_SIZE as u32 {
+                panic!("Mismatched read sizes for `vm_read` (expected {}, got {})", VM_PAGE_SIZE, read_len)
             }
 
             let read_buf = unsafe { slice::from_raw_parts(read_ptr, read_len as usize) };
 
-            let offset = addr - page_addr as usize;
-            let len = buf.len();
-            buf.copy_from_slice(&read_buf[offset..(offset + len)]);
+            buf.extend_from_slice(read_buf);
```
kamalmarhubi: does this still result in "polluting" vm space as new mappings are made by `vm_read`? if so, please add a FIXME mentioning it, so we know it still needs to be sorted out.

src/lib.rs
kamalmarhubi: oh I see, this is directly adapting from xnu. feel free to ignore.

kamalmarhubi:
Nice! I have a couple of tiny comments but this looks great.

For later, I figure that the "right" way to fix the `vm_read` will be to use `vm_read_overwrite` which writes to a user-specified address instead of dynamically creating a mapping for the copy. It may be messy because the destination might also have to be page-aligned, and there's no way to allocate on a page boundary in stable Rust. Only thing I can see doing is create a 2x page size buffer and get a page-aligned slice of it. Anyway, getting ahead of things there for now. :-)

luser:
FYI I factored the process memory reading bits of this project out to a standalone crate: https://github.com/luser/read-process-memory

I don't have the automated tests running on OS X yet (haven't worked around the `task_for_pid` nonsense), but I did some manual testing and it looked like on my system (10.11.6) vm_read worked fine with a size > 4096. Am I missing something?

I've got a PR here to use that crate for ruby-stacktrace: https://github.com/jvns/ruby-stacktrace/pull/14

kamalmarhubi:
@luser does your change in #14 cover unaligned reading on OS X? If so I'll close this PR. If not, it's probably worth adapting @dirk's work the `read-process-memory` crate. Either way drop a note here so we know what to do!

kamalmarhubi:
Ugh reading comprehension. Saw your note about the size. I don't have an OS X system handy to test on, so I can't investigate.

luser:
I did manual testing with both >4k reads and unaligned reads and they seemed to work with the existing code on my machine. I don't know if that's because of changes in the kernel, or if there's a scenario that @dirk tested that I missed. If this code is needed for OS X I'd be happy to take it into `read-process-memory`.

kamalmarhubi:
Ok. I'm going to close this out. If a user finds it's not working properly we can revisit.

dirk:
> I did manual testing with both >4k reads and unaligned reads and they seemed to work with the existing code on my machine. I don't know if that's because of changes in the kernel, or if there's a scenario that @dirk tested that I missed. If this code is needed for OS X I'd be happy to take it into read-process-memory.

@luser: Honestly this patch was just to have it do The Right Thing like all the other code I'd seen using `vm_read` (mostly looking at the XNU source itself as well as the JVM HotSpot's debugger code). If the cases you tested work then I'd call the existing implementation "Right Enough". 😉

PR #11, opened 2016-09-13
typo fix for url of gdb python script that was a basis of this rust port

fche:


kamalmarhubi:
What an odd typo! Thanks :-)

PR #13, opened 2016-10-11
License add

### LINKED ISSUE #12: License?
Hey there! I love the idea of this tool, and was toying with the idea of doing something similar for other languages. I'd love to reuse some of your code, but you don't have a license on it currently. Would you be amenable to adding an open source license so I can reuse it? I spent a little time fiddling yesterday and extracted the "read process memory" bits into a standalone crate (and even added a Windows implementation), so if you add a license I can publish that to crates.io and submit a PR for this project to use it. :)

petems:
Adds MIT license 

Closes #12

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ name = "ruby-stacktrace"
 version = "0.1.0"
 authors = ["Julia Evans <julia@jvns.ca>"]
 
+license = "MIT"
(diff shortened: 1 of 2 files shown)


petems:
If put both @jvns and @kamalmarhubi under the MIT license text, wasn't sure because the crate only has @jvns. Can amend if needed! 👍

PR #14, opened 2016-10-12
Use the new read-process-memory crate

luser:
Hey! I wanted to use your process-memory-reading code for another project, so I split it out to a standalone crate:
https://github.com/luser/read-process-memory
https://crates.io/crates/read-process-memory

Here's a PR to remove that code from this project and make it use that crate instead. I even added a Windows implementation to that crate if you ever get strangely motivated to make ruby-stacktrace work on Windows. :)

While I was removing code I removed all but one of your `unsafe` blocks! I think some of them were probably just leftovers from previous hacking.


jvns:
Thanks so much for this!

PR #48, opened 2018-01-21
tests: Move core dumps to a separate crate

kamalmarhubi:
Keep main repo size down post-squash.


jvns:
oh, this is clever. Nice!

PR #57, opened 2018-01-22
Specify sleep time in nanoseconds instead of ms

jvns:
Previously we were always sleeping for an integer number of milliseconds between sampler runs. This isn't very accurate -- if someone wants to sample (say) 172 times per second, we should sleep for 5.8 milliseconds, not 6 milliseconds.

This changes the `record` function to use `Duration::new(0, num_nanoseconds)` instead of `Duration::from_millis(num_millis)` so we can sleep more accurately.

 cc @liaden

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -48,6 +48,7 @@ pub mod copy;
 pub mod ruby_version;
 
 const FLAMEGRAPH_SCRIPT: &'static [u8] = include_bytes!("../vendor/flamegraph/flamegraph.pl");
+const BILLION: u32 = 1000 * 1000 * 1000; // for nanosleep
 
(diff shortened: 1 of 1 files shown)


liaden:
Would it be worthwhile to use https://github.com/brayniac/shuteye for nanosleep since it already is wrapping libc and handling different OSes?

jvns:
I was wondering about that! It's not clear to me what advantage that crate has over `std::thread::sleep` from the standard library.

liaden:
Looking at the docs for https://doc.rust-lang.org/beta/std/thread/fn.sleep.html#platform-behavior, I am unsure since it does allow nanoseconds for some platforms.

PR #58, opened 2018-01-22
Record for fixed length of time

liaden:
This  checks the time elapsed on every collected stacktrace to see if we are have hit the users specified duration threshold if they set the duration.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -65,7 +65,7 @@ use Target::*;
 #[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
 enum SubCmd {
     /// Record `target`, writing output `output`.
-    Record { target: Target, out_path: PathBuf, sample_rate: u32  },
+    Record { target: Target, out_path: PathBuf, sample_rate: u32, maybe_duration: Option<std::time::Duration>},
     /// Capture and print a stacktrace snapshot of process `pid`.
     Snapshot {pid: pid_t},
 }
@@ -85,7 +85,7 @@ fn do_main() -> Result<(), Error> {
 
     match args.cmd {
         Snapshot { pid } => snapshot(pid),
-        Record { target, out_path, sample_rate } => {
+        Record { target, out_path, sample_rate, maybe_duration } => {
             let (pid, spawned) = match target {
                 Pid {pid} => (pid, false),
                 Subprocess { prog, args } => (Command::new(prog)
@@ -93,7 +93,7 @@ fn do_main() -> Result<(), Error> {
                     .spawn()?.id() as pid_t, true)
             };
 
-            record(&out_path, pid, sample_rate, spawned)
+            record(&out_path, pid, sample_rate, maybe_duration, spawned)
         },
     }
 }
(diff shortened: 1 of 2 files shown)


jvns:
This is great! Do you think it makes sense for `--duration` to conflict with the `cmd` command line option (using `conflicts_with`)? 

I think the meaning of `rbspy record --duration 10 ruby script.rb` is kind of unclear (is the goal to run `ruby script.rb` but then kill it after 10 seconds?) and it might be easier to just disallow that combination

src/main.rs
```diff
@@ -121,13 +121,17 @@ fn snapshot(pid: pid_t) -> Result<(), Error> {
     Ok(())
 }
 
-fn record(output_filename: &Path, pid: pid_t, sample_rate: u32, is_subcommand: bool) -> Result<(), Error> {
+fn record(output_filename: &Path, pid: pid_t, sample_rate: u32, duration: u64, is_subcommand: bool) -> Result<(), Error> {
```
jvns: it would be more idiomatic Rust for duration to be an `Option<u64>`, so that duration = None means that duration is unset instead of duration = 0

jvns: it could actually even be an `Option<Duration>`!

liaden: Sweet, will update.

liaden:
I am fine with going either way on `--duration` conflicting with the `cmd` option. Naively, it feels like it is like an automated `ctrl-c`, but I don't see a use case for supporting the two together at this point.

src/main.rs
```diff
@@ -174,6 +177,14 @@ fn record(output_filename: &Path, pid: pid_t, sample_rate: u32, is_subcommand: b
                 }
             }
         }
+        match stop_time {
```
jvns: idiomatic rust note: you can make this a little shorter with an `if let`:
```
if let Some(time) = stop_time {
     ...
}

File: src/main.rs [hunk 3]
```diff
@@ -121,13 +121,16 @@ fn snapshot(pid: pid_t) -> Result<(), Error> {
...
+fn record(output_filename: &Path, pid: pid_t, sample_rate: u32, duration: Option<std::time::Duration>, is_subcommand: bool) -> Result<(), Error> {
     // This gets a stack trace and then just prints it out
     // in a format that Brendan Gregg's stackcollapse.pl script understands
     let getter = initialize::initialize(pid)?;
 
     eprintln!("Recording data to {}", output_filename.display());
-    eprintln!("Press Ctrl+C to stop");
+    let stop_time = match duration {
+        None => { eprintln!("Press Ctrl+C to stop"); None },
+        Some(dur) => { Some(std::time::Instant::now() + dur) }
```
liaden: Is there a rust pattern for naming optionals vs naming the unwrapped value?

liaden: That is a good naming practice and keeping things consistent is good.

src/main.rs
liaden: The first is terser while the second is likely nicer when there are 3 or more types to match against.

src/main.rs
```diff
@@ -174,6 +177,19 @@ fn record(output_filename: &Path, pid: pid_t, sample_rate: u32, is_subcommand: b
...
+            }
+        }
+        //match stop_time {
+        //    None => {},
+        //    Some(time) => {
+        //       if std::time::Instant::now() > time {
+        //           return Ok(());
+        //       }
+        //    }
+        //}
```
jvns: can you just delete this commented out bit?

src/main.rs
```diff
@@ -318,6 +329,11 @@ impl Args {
             },
             ("record", Some(submatches)) => {
                 let out_path = output_filename(&std::env::var("HOME")?, submatches.value_of("file"))?;
+                let maybe_duration = match value_t!(submatches, "duration", u64).unwrap_or(0) {
```
jvns: one last note -- instead of doing `.unwrap_or`, you could match against the option returned by `value_t!` directly (`None => None...`)

liaden: I had left the `unwrap_or` in place because I kept getting regardless of using `=> { None }` or `=> None` :
```
error[E0308]: mismatched types
   --> src/main.rs:333:21
    |
333 |                     None =>  None ,
    |                     ^^^^ expected enum `std::result::Result`, found enum `std::option::Option`
    |
    = note: expected type `std::result::Result<u64, clap::Error>`
               found type `std::option::Option<_>`
```

jvns: oops, that's right, this is a Result and not an Option (https://doc.rust-lang.org/std/result/index.html). the right way to match that is

```
match result_thing {
    Err(_) => None,
    Ok(integer) => ...
}
```

jvns: actually  looking at the [docs for the `value_t` macro](https://docs.rs/clap/2.9.3/clap/macro.value_t!.html) more carefully, it looks like we can actually differentiate between "the argument was missing" and "invalid argument" using `value_t`: you can match against this ErrorKind to differentiate https://kbknapp.github.io/clap-rs/clap/enum.ErrorKind.html

```
Err(ErrorKind::EmptyValue) => None,
Err(x) => return Err(x.into(), // the `x.into()` is an annoying thing we have to do because of Rust
Ok(x) => ...
```

PR #63, opened 2018-01-24
Validate user's filename

liaden:
The filename writes out the raw stacks, and is unable to write the flamegraph if the user specifies a file with the .svg extension.


jvns:
good catch, thanks!

Issue #64, opened 2018-01-25
'allocator memory exhausted' error when spying on a Rails process


ragesoss:
I'm giving this a spin by trying to profile bits of [my Rails app](https://github.com/WikiEducationFoundation/WikiEduDashboard).

Usually, when I load a page on the spied-upon rails server, rbspy exits immediately like this:

```
$ sudo rbspy record --pid 32471
Recording data to /root/.cache/rbspy/records/rbspy-2018-01-25-tw0bEjjEZq.txt
Press Ctrl+C to stop
fatal runtime error: allocator memory exhausted
Illegal instruction
```
It works for lighter page loads, like json endpoints that don't load or serve all that much data, but for typical tasks in my app that I'd like to profile, I haven't been able to get around this error.

liaden:
Which endpoints are having an issue? Is it problematic in dev as well as prod?

ragesoss:
@liaden I've only tried it in dev. It seems to be inconsistent, but more likely to crash on large sql queries and/or ones that send a lot of data.

Also, sometimes after it's crashed once, it will crash again immediately if I try to spy again on the same still-running rails server process (even if I don't load any pages).

I can try to provide more details, but I'm not sure which ones are likely to be useful.

jvns:
this is an interesting issue, thanks for reporting! what Ruby version are you using? I made a PR with a mitigation in #65, which should help a little bit -- will make a new release from that so you can try it.

ragesoss:
Ruby 2.5.0. :)

jvns:
@ragesoss can you try with the latest release and see if it works any better? https://github.com/rbspy/rbspy/releases/tag/v0.1.3

If there are still a lot of errors, it would be great if you could run rbspy as  `env RUST_LOG=debug rbspy` and paste the output somewhere.

ragesoss:
Yes, it works much better! If I spy the already-started rails server, it doesn't crash now.

With or without `env RUST_LOG=debug`, I don't see any errors or logging when I spy the process and do lots of page loading on the app.

The flamegraphs look like they might only be capturing the first page load or so. But it looks like logging worked; it captured 10 MB in the text file, compared to mostly less than 1MB before crashing before.

![activity5](https://user-images.githubusercontent.com/848483/35368268-6164dd1c-0137-11e8-8047-e3bafdfcb9de.png)

liaden:
[...]

PR #65, opened 2018-01-25
Return an error if we request more than 1Mb of memory in copy_address_raw

jvns:
there's an issue (#64) where if some memory is incorrect, we can end up passing in an invalid length to `copy_address_raw` (length = some huge number), then allocating huge amounts of memory in `copy_address_raw` and the program crashes. This PR doesn't fix the underlying issue (why is the memory in an invalid state?), but it should avoid allocating huge amounts of memory and instead log an error and continue. The choice of 1MB is arbitrary.

In the future we should be more disciplined about logging more detailed output than this when there's an error (so it's possible to ask users for the debug output)

### DIFF
--- a/src/copy.rs
+++ b/src/copy.rs
@@ -5,6 +5,8 @@ use read_process_memory::*;
  * Utility functions for copying memory out of a process
  */
 
+const MAX_COPY_LENGTH: usize = 1000000;
+
 #[derive(Fail, Debug)]
 pub enum MemoryCopyError {
     #[fail(display = "Permission denied when reading from process. Try again with sudo?")]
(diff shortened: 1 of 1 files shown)


liaden:
Is this a case of the callstack being too deep?

jvns:
I don't know what the root cause is! I'm going to make another PR that prints more debug output when there's an error.

src/copy.rs
```diff
@@ -12,6 +12,7 @@ pub enum MemoryCopyError {
     #[fail(display = "Failed to copy memory address {:x}", _0)] Io(usize, #[cause] std::io::Error),
     #[fail(display = "Process isn't running")] ProcessEnded,
     #[fail(display = "Other")] Other,
+    #[fail(display = "Too much memory requested when copying")] InvalidLength,
```
kamalmarhubi: I'd make the name be `RequestTooLarge` or something more indicative of the actual issue. Also, take a `usize` parameter and include it in the display message.

src/copy.rs
```diff
@@ -36,6 +37,9 @@ where
     T: CopyAddress,
 {
     debug!("copy_address_raw: addr: {:x}", addr as usize);
+    if length > 1000000 {
```
kamalmarhubi: make this a constant up top?

PR #66, opened 2018-01-25
Only print debug output when there are errors

jvns:
This makes the debugging output more organized/useful -- the main time I think debugging output will be useful is when there's an error getting a stack trace. So this moves most of the debug printing to the case where there's actually an error.

This should make it easier to ask users "hey, if you run into an error, rerun it with `env RUST_LOG=debug` and send us the output of that"

This prints out the cfp, iseq, and thread structs, as well as the rstring struct, which I'm hoping will be a good start.

### DIFF
--- a/src/ruby_version.rs
+++ b/src/ruby_version.rs
@@ -123,31 +123,28 @@ macro_rules! get_stack_trace(
             ruby_current_thread_address_location: usize,
             source: &T,
             ) -> Result<Vec<StackFrame>, MemoryCopyError> where T: CopyAddress{
-            debug!(
-                "current address location: {:x}",
-                ruby_current_thread_address_location
-                );
             let current_thread_addr: usize =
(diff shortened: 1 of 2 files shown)


src/ruby_version.rs
```diff
@@ -252,7 +247,15 @@ macro_rules! get_ruby_string(
                     unsafe {
                         let addr = rstring.as_.heap.ptr as usize;
                         let len = rstring.as_.heap.len as usize;
-                        copy_address_raw(addr as usize, len, source)?
+                        let result = copy_address_raw(addr as usize, len, source);
+                        match result {
+                            Err(x) => {
+                                debug!("Error: Failed to get ruby string.");
+                                debug!("rstring: {:?}, addr: {}, len: {}", rstring, addr, len);
```
kamalmarhubi: not clear why these are two different log lines.

PR #68, opened 2018-01-25
Add support for profiling processes in Linux containers

### LINKED ISSUE #67: trace processes that run in docker-container.
Hi - love rbspy already - big thanx. I can see this becoming a very important tool for all the ruby people out there!

I run quite a few things in docker container and love to trace them _without_ having to enter the container (i would also run the container privileged I believe to use rbspy within the container). 

To reproduce:
docker-host> docker run ruby ruby -e "sleep(500)"

(in different shell)
docker-host> ps auxww |grep "ruby -e sleep"
root     27179  0.0  0.0  38008  7460 ?        Ssl  16:28   0:00 ruby -e sleep(500)
docker-host> sudo ./rbspy snapshot --pid 27179
Error. Causes:
- Couldn't determine Ruby version
- Couldn't open ELF file: /usr/local/bin/ruby

This is do to the fact that ruby is of course installed in the container somewhere in the depth of /var/lib/docker ...

jvns:
Fixes #67.

This adds support for profiling Ruby processes running in Linux containers! 

The only reason rbspy **didn't** work with containers was -- there's a small part at the beginning of the program where it reads 1 or 2 binaries from the program's memory maps (the Ruby binary, and sometimes another dynamically linked library). 

It's also worth making "work with containers" more precise -- the real issue was that the target process was in a **different mount namespace** (/usr/bin/ruby1.9.1 in the target process didn't mean the same thing as /usr/bin/ruby1.9.1 in rbspy's process). 

So to make rbspy work with containerized processes, we just need to switch to the target process's mount namespace before reading those two ELF binaries. Then after reading the ELF binaries we switch back to our mount namespace right away so that we can write output to the filesystem in the right place. Pretty simple!

Originally I wanted to create a separate thread to do the namespace-switching juggle but it turns out that you can't run `setns` if your program has multiple threads. So switching & then switching back it is.

### DIFF
--- a/src/address_finder.rs
+++ b/src/address_finder.rs
@@ -71,10 +71,15 @@ mod os_impl {
     use elf;
     use proc_maps::*;
     use failure::Error;
+    use failure::ResultExt;
     use libc::pid_t;
+    use libc;
     use std;
+    use std::fs;
+    use std::os::linux::fs::MetadataExt;
     use address_finder::AddressFinderError;
     use read_process_memory::*;
+    use std::os::unix::prelude::*;
 
     pub fn current_thread_address(
         pid: pid_t,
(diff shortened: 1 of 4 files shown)


baloo:
I just realized that rbspy was opening `/usr/bin/ruby1.9.1` directly instead of using `/proc/PID/exe`. The first one might have been replaced (by apt/yum/...) by the time rbspy is started and elf could be all wrong.

jvns:
this is true! Hmm. Using `/proc/$PID/exe` isn't actually enough to prevent this program because we also the dynamically linked library (`libruby-whatever.so`).

Maybe it's better to read the binaries we need out of the process's memory maps instead of trying to read them from the filesystem?

baloo:
Most likely yes, but some pages will be rewriten by the dynamic loader. I don't know if `elf` crate is fine with that (it should be).

src/address_finder.rs
```diff
@@ -118,6 +123,50 @@ mod os_impl {
         }
     }
 
+    fn switch_ns(fd: RawFd) -> Result<(), Error> {
```
kamalmarhubi: to fix the fd lifetime issue, take it by reference. to be extra careful,

```rust
fn switch_ns<F: AsRawFd>(fd: &F) -> Result<(), Error> {
    let fd = fd.as_raw_fd();
    // ...
}
```

this means you need to have a thing that "is" a file descriptor,and that it has to outlive the call to `switch_ns`. you'd pass in the file instead of the fd below.

src/address_finder.rs
```diff
@@ -118,6 +123,50 @@ mod os_impl {
...
+        } else {
+            // switch mount namespace and then switch back after getting the program info
+            // We need to save the fd of the current mount namespace so we can switch back
+            let f = fs::File::open(other_proc_mnt)?;
+            let new_ns_fd = f.as_raw_fd();
+            let f = fs::File::open(self_proc_mnt)?;
+            let old_ns_fd = f.as_raw_fd();
+            switch_ns(new_ns_fd)?;
+            let proginfo = get_program_info(pid, all_maps);
+            switch_ns(old_ns_fd)?;
```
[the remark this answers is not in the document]
kamalmarhubi: the easy way to fix this is to use `defer!` from scope_guard. the slightly less easy way is to make your own RAII guard type for switched ns.

```rust
defer!(switch_ns(&old_ns).expect("failed to switch to original namespace"));
```

or `defer_on_unwind` if you want to not unwrap on non-failing exit.

src/address_finder.rs
```diff
@@ -71,17 +71,22 @@ mod os_impl {
...
     use read_process_memory::*;
+    use std::os::unix::prelude::*;
 
     pub fn current_thread_address(
         pid: pid_t,
         version: &str,
         is_maybe_thread: Box<Fn(usize, &ProcessHandle, &Vec<MapRange>) -> bool>,
     ) -> Result<usize, Error> {
-        let proginfo = &get_program_info(pid)?;
+        let proginfo = &get_program_info_wrapper(pid)?;
```
kamalmarhubi: I'd leave this named as is, and call the other one something like `_get_program_info` or `inner` or something.

jvns:
I'm going to leave this the way it is for now, but I think switching to using `/proc/$PID/exe` is a good idea for a separate PR. I went down the "can we read dynamically loaded libraries from the target process's memory?" rabbit hole a little bit and didn't get it to work.

PR #71, opened 2018-01-27
Write flamegraph when we record for a duration

jvns:
Just noticed that when we were recording for a duration, it wouldn't write a flamegraph at the end. This should fix that! It felt like there were a lot of calls to `write_flamegraph` to keep track of so I thought it would be better to restructure the code a tiny bit to ensure that we always call `write_flamegraph` after we finish recording.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -125,6 +125,12 @@ fn snapshot(pid: pid_t) -> Result<(), Error> {
 }
 
 fn record(output_filename: &Path, pid: pid_t, sample_rate: u32, maybe_duration: Option<std::time::Duration>, is_subcommand: bool) -> Result<(), Error> {
+    do_record(output_filename, pid, sample_rate, maybe_duration, is_subcommand)?;
+    write_flamegraph(output_filename)?;
+    Ok(())
+}
+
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -125,6 +125,12 @@ fn snapshot(pid: pid_t) -> Result<(), Error> {
 }
 
 fn record(output_filename: &Path, pid: pid_t, sample_rate: u32, maybe_duration: Option<std::time::Duration>, is_subcommand: bool) -> Result<(), Error> {
+    do_record(output_filename, pid, sample_rate, maybe_duration, is_subcommand)?;
+    write_flamegraph(output_filename)?;
+    Ok(())
+}
+
+fn do_record(output_filename: &Path, pid: pid_t, sample_rate: u32, maybe_duration: Option<std::time::Duration>, is_subcommand: bool) -> Result<(), Error> {
```
liaden: Could the name be more descriptive than `do_record`? Maybe `generate_raw_data`?

liaden:
Following all the logging was a bit much when I was tracking down how to fix the `--file myfile.svg` bug so definitely worth the restructure.

jvns:
@vasi fixed this as an awesome sideeffect of #75 (thanks!) so closing :)

PR #74, opened 2018-01-27
Simplify container support

jvns:
This reverts the previous commit adding container support and implements a much simpler approach instead! 

It turns out that to read a file from a process's mount namespace, you can just read it from `/proc/PID/root/<filename>`. I like this because we can just always implement the same approach whether the target process is running in a container or not!

### DIFF
--- a/src/address_finder.rs
+++ b/src/address_finder.rs
@@ -71,15 +71,10 @@ mod os_impl {
     use elf;
     use proc_maps::*;
     use failure::Error;
-    use failure::ResultExt;
     use libc::pid_t;
-    use libc;
     use std;
-    use std::fs;
-    use std::os::linux::fs::MetadataExt;
     use address_finder::AddressFinderError;
     use read_process_memory::*;
-    use std::os::unix::prelude::*;
 
     pub fn current_thread_address(
         pid: pid_t,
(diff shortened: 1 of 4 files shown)


src/address_finder.rs
```diff
@@ -238,6 +238,17 @@ mod os_impl {
...
 
+    fn open_elf_file(pid: pid_t, map: &MapRange) -> Result<elf::File, Error> {
+        // Read binaries from `/proc/PID/root` because the target process might be in a different
+        // mount namespace. /proc/PID/root is the view of the filesystem that the target process
+        // has. (see the proc man page for more)
+        // So we read /usr/bin/ruby from /proc/PID/root/usr/bin/ruby
+        let map_path = map.pathname.as_ref().expect("map's pathname shouldn't be None");
+        let elf_path = format!("/proc/{}/root{}", pid, map_path);
+        elf::File::open_path(&elf_path)
+            .map_err(|_| format_err!("Couldn't open ELF file: {:?}", elf_path))
```
kamalmarhubi: `.context(...)?`

jvns: > error[E0599]: no method named `context` found for type `std::result::Result<elf::File, elf::ParseError>` in the current scope

jvns:
```
error[E0599]: no method named `context` found for type `std::result::Result<elf::File, elf::ParseError>` in the current scope
   --> src/address_finder.rs:249:41
    |
249 |         elf::File::open_path(&elf_path).context(format_err!("Couldn't open ELF file: {:?}", elf_path))
    |                                         ^^^^^^^
    |
    = note: the method `context` exists but the following trait bounds were not satisfied:
            `std::result::Result<elf::File, elf::ParseError> : failure::ResultExt<_, _>`
```

PR #75, opened 2018-01-28
Add callgrind output

vasi:
## Why?

Flamegraphs are awesome! But sometimes there are better ways to visualize your program.

For example, here's the [dumbest fibonacci calculator ever](https://gist.github.com/vasi/ae3b25ce7bfd7540be6c3970b9a0a4f6), and the flamegraph rbspy generates:

image

It's super hard to tell what's wrong. Even though `fib0` is being called a gazillion times and doing loads of work, it doesn't really show up. On the other hand, I can run.:

```
sudo ./target/debug/rbspy record -f ../callgrind.out --format callgrind ./fib.rb 19
```

And then open `callgrind.out` with [qcachegrind](https://kcachegrind.github.io/html/Home.html), I get some really useful output:

image

Clearly "block in fib0" has something wrong!

### Bonus!

* qcachegrind can do some nice visualizations:
image
* callgrind files are much smaller than flamegraphs—easier to collect for future analysis or share with friends

## What's changed

* A new argument `--format=` allows specifying either `flamegraph` (default) or `callgrind`
* `callgrind.rs` contains code to aggregate stack traces, and to write the callgrind file format
* `output.rs` takes over the flamegraph code, and adds some glue for callgrind
* `record()` now takes an `Outputter` as an argument, and uses that to do all the real work. I think this overall makes it cleaner
    * It now has fewer exit points—all success-like ones (ctrl-c, program finished, duration expired) go through the same exit.
    * This also fixes a bug where duration-expired wouldn't cause the flamegraph to be written

## Caveats

* The "number of times each function is called" is inaccurate. This is because if two samples have identical call stacks, we have no way to tell whether there have been intermediate changes. Maybe there's a way to get this info from Ruby.
* Other formats (pprof, perf, gprof...) might be more useful in some situations. But callgrind is [well-specified](http://valgrind.org/docs/manual/cl-format.html) and text-based, so much easier to implement.
* My Rust is pretty awful, a lot of StackFrames are getting copied for no good reason. Not sure if that's fixable!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -55,6 +55,15 @@ enum Target {
 }
 use Target::*;
 
+// Formats we can write to
+arg_enum!{
+    #[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+    pub enum OutputFormat {
+        Flamegraph,
+        Callgrind,
+    }
+}
+
 /// Subcommand.
 #[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
 enum SubCmd {
@@ -64,6 +73,7 @@ enum SubCmd {
         out_path: PathBuf,
         sample_rate: u32,
         maybe_duration: Option<std::time::Duration>,
+        format: OutputFormat,
     },
     /// Capture and print a stacktrace snapshot of process `pid`.
     Snapshot { pid: pid_t },
@@ -89,6 +99,7 @@ fn do_main() -> Result<(), Error> {
             out_path,
             sample_rate,
             maybe_duration,
+            format,
         } => {
             let (pid, spawned) = match target {
                 Pid { pid } => (pid, false),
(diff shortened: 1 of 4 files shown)


vasi:
Uh not sure why these commits are coming from my work account, will fix.

src/main.rs
```diff
@@ -149,112 +154,73 @@ fn record(
...
 
-    let mut output = std::fs::File::create(&output_filename).context(format!(
+    let mut out_file = File::create(&out_path).context(format!(
         "Failed to create output file {}",
-        &output_filename.display()
+        &out_path.display()
     ))?;
-    loop {
-        total.fetch_add(1, Ordering::Relaxed);
+    while !done.load(Ordering::Relaxed) {
```
jvns: This is a much cleaner way of doing Ctrl+C handling, thank you!

jvns:
Thanks so much for this PR! This is super useful, and I love that you cleaned up a few other things that had been bothering me -- the new way of doing ctrl+c handling in this PR is way better. The new `Outputter` trait is nice too.

I'm planning to do some performance work later anyway so I'm not too worried about any extra copies this causes.

PR #78, opened 2018-01-30
Display StackFrames in the Ruby Kernel#caller format.

mvidner:
Example:

If we take the "Dumbest Fibonacci ever" program used in the Callgrind
PR, https://gist.github.com/vasi/ae3b25ce7bfd7540be6c3970b9a0a4f6
it looks like this:

```
$ ruby /tmp/fib.rb 22 &
$ sudo ./target/debug/rbspy snapshot --pid $(pgrep -f fib.rb)
/tmp/fib.rb:163:in `<main>'
/tmp/fib.rb:159:in `fib'
/tmp/fib.rb:113:in `fib22'
/tmp/fib.rb:108:in `fib21'
/tmp/fib.rb:103:in `fib20'
/tmp/fib.rb:98:in `fib19'
/tmp/fib.rb:88:in `fib17'
/tmp/fib.rb:83:in `fib16'
/tmp/fib.rb:78:in `fib15'
/tmp/fib.rb:73:in `fib14'
/tmp/fib.rb:68:in `fib13'
/tmp/fib.rb:63:in `fib12'
/tmp/fib.rb:53:in `fib10'
/tmp/fib.rb:48:in `fib9'
/tmp/fib.rb:38:in `fib7'
/tmp/fib.rb:33:in `fib6'
/tmp/fib.rb:23:in `fib4'
/tmp/fib.rb:13:in `fib2'
/tmp/fib.rb:4:in `fib0'
/tmp/fib.rb:4:in `block in fib0'
```

This changes the default output format. If something relies on
the original format I will be happy to add command line switches
for the formats.

My biggest motivation is to use the colon-delimited filename:line
format because it is common in compiler output and many tools
know how to open such locations in code editors.

I actually hate the backtick quote in the output because it messes up 
syntax highlighters, but I think it is valuable to have exact 
compatibility of the output.

### DIFF
--- a/src/initialize.rs
+++ b/src/initialize.rs
@@ -83,7 +83,7 @@ pub struct StackTraceGetter {
 
 impl fmt::Display for StackFrame {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{} - {} line {}", self.name, self.path(), self.lineno)
+        write!(f, "{1}:{2}:in `{0}'", self.name, self.path(), self.lineno)
     }
 }
 


jvns:
@mvidner thanks for this! Can you give an example of a tool that knows how to open locations in code editors? Knowing what tools do that will make it easier for me to think about this.

The motivation for using the current format is mostly that when displaying a flamegraph it makes more sense to put the function name first. (because there's limited space). But when printing to the terminal the order doesn't matter so much, and we can definitely use different formats for flamegraphs vs the terminal.

mvidner:
Ha, I got a bit ahead of myself with "many tools", it's mostly just my [private `e` alias](https://bitbucket.org/mvidner/dotfiles/src/f077d51b43d65198d066ef9bf2fd5e9e0a7ca8d0/.bash_aliases?at=master&fileviewer=file-view-default#.bash_aliases-4) but the format it consumes is common to
- grep -n # --line-number
- GCC errors/warnings
- [The GNU coding standards](https://www.gnu.org/prep/standards/standards.html#Errors)

For flamegraphs, the original format totally makes sense.

liaden:
For what it is worth, the ruby caller format may be a better default for when doing a snapshot?

acj:
Closing due to age. I'd be happy to consider a fresh version of this if it's still needed.

PR #79, opened 2018-01-31
Make sure we sample the correct number of times per second

jvns:
Previously we were sampling 100 times per second by doing a sleep (1 billion nanoseconds / 100). This isn't totally accurate because of course *taking* a sample takes a nonzero amount of time.  This introduced a skew on my laptop of about 3%.

This fixes that by using a smarter strategy for sleeping -- instead of sleeping for X nanoseconds, instead we calculate the expected next sample time and sleep until that time.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -36,6 +36,7 @@ use std::env;
 use std::process::Command;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
+use std::time::Instant;
 
 pub mod proc_maps;
 pub mod address_finder;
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -151,6 +152,34 @@ impl OutputFormat {
...
 
+// This SampleTime struct helps us sample on a regular schedule ("exactly" 100 times per second, if
+// the sample rate is 100).
+// What we do is -- when doing the 1234th sample, we calculate the exact time the 1234th sample
+// should happen at, which is (start time + nanos_between_samples * 1234) and then sleep until that
+// time
+struct SampleTime {
+    start_time: Instant,
+    nanos_between_samples: u64,
+    num_samples: u64,
```
kamalmarhubi: what do the fields mean?

jvns:
todo: if someone tries to sample at a rate faster than the fastest possible rate, we should crash and ask you to sample at a slower rate.

Issue #82, opened 2018-01-31
linux: thread 'main' panicked at 'attempt to subtract with overflow'


parkr:
Hi! I just pulled & re-compiled at 5f7b3615277978d8e50dc335cdfdbb4910e8ddc9 and got a funky error! Looks like something funky with `rbspy::SampleTime::get_sleep_time` – maybe related to #79?

```text
~/rbspy$ git rev-parse HEAD
5f7b3615277978d8e50dc335cdfdbb4910e8ddc9
~/massive-jekyll-site$ rustc --version
rustc 1.23.0 (766bd11c8 2018-01-01)
~/rbspy$ which ruby
/usr/bin/ruby
~/massive-jekyll-site$ ruby --version
ruby 2.3.3p222 (2016-11-21) [x86_64-linux-gnu]
~/massive-jekyll-site$ RUST_BACKTRACE=1 ~/rbspy/target/debug/rbspy record /home/parker/jekyll/exe/jekyll build
Recording data to /home/parker/.cache/rbspy/records/rbspy-2018-01-31-Jgn67RAD6i.txt
Press Ctrl+C to stop
Configuration file: none
            Source: /home/parker/massive-jekyll-site
       Destination: /home/parker/massive-jekyll-site/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
thread 'main' panicked at 'attempt to subtract with overflow', src/main.rs:224:9
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:68
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:57
             at /checkout/src/libstd/panicking.rs:381
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:397
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:577
   5: std::panicking::begin_panic
             at /checkout/src/libstd/panicking.rs:538
   6: std::panicking::begin_panic_fmt
             at /checkout/src/libstd/panicking.rs:522
   7: rust_begin_unwind
             at /checkout/src/libstd/panicking.rs:498
   8: core::panicking::panic_fmt
             at /checkout/src/libcore/panicking.rs:71
[...]

PR #83, opened 2018-01-31
Don't crash if we fall behind our expected sample time

jvns:
Fixes https://github.com/rbspy/rbspy/issues/82

#79 made sampling more accurate (yay!) but introduced an issue where rbspy would crash if it fell behind its expected sampling times (for example because someone is sampling at too high of a rate). This changes it so that rbspy will warn that results may be inaccurate and that it's dropping a sample, but it won't crash.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -217,11 +217,18 @@ impl SampleTime {
         }
     }
 
-    pub fn get_sleep_time(&mut self) -> u32 {
+    pub fn get_sleep_time(&mut self) -> Result<u32, u32> {
+        // Returns either the amount of time to sleep (Ok(x)) until next sample time or an error of
+        // how far we're behind if we're behind the expected next sample time
         self.num_samples += 1;
         let elapsed = self.start_time.elapsed();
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -289,8 +296,11 @@ fn record(
             }
         }
         // Sleep until the next expected sample time
-        let sleep_time = sample_time.get_sleep_time();
-        std::thread::sleep(std::time::Duration::new(0, sleep_time));
+        //
+        match sample_time.get_sleep_time(){
+            Ok(sleep_time) => {std::thread::sleep(std::time::Duration::new(0, sleep_time));},
+            Err(behind_time) => {eprintln!("Behind expected sample time by {} nanoseconds, results may be inaccurate. Try sampling at a lower rate", behind_time);},
```
parkr: The one thing I might find useful here is if this told me how to sample! What is the current rate, and how to I set a lower rate?

jvns:
thanks for the comment, i made the error message a little more detailed!

Issue #80, opened 2018-01-31
OS error 17 when running rbspy on mac


jvns:
@parkr reported having this issue on Mac (High Sierra) when trying to run rbspy. I haven't been able to reproduce this yet but creating an issue to track it!

```
~/code/rbspy#mac-support$ rm -rf target && cargo build
~/code/rbspy#mac-support$ sudo --preserve-env ./target/debug/rbspy record ruby ./ci/ruby-programs/infinite.rb Dropping permissions: running Ruby command as user parkr
Error. Causes:
- Couldn't determine Ruby version
- Couldn't connect to PID
- File exists (os error 17)

~/code/rbspy#mac-support$ rbenv version
2.4.3 (set by /Users/parkr/code/rbspy/.ruby-version)
~/code/rbspy#mac-support$ sudo --preserve-env env RUST_LOG=debug RUST_BACKTRACE=1 ./target/debug/rbspy record ruby ./ci/ruby-programs/infinite.rb
Dropping permissions: running Ruby command as user parkr
Error. Causes:
- Couldn't determine Ruby version
- Couldn't connect to PID
- File exists (os error 17)
stack backtrace:
   0:        0x10fd4ab2c - backtrace::backtrace::trace<closure>
                        at /Users/parkr/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.5/src/backtrace/mod.rs:42
   1:        0x10fd4210f - backtrace::capture::Backtrace::new_unresolved::h22a3194b78596c3b
                        at /Users/parkr/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.5/src/capture.rs:88
   2:        0x10fd3ffc1 - failure::backtrace::internal::{{impl}}::new
                        at /Users/parkr/.cargo/registry/src/github.com-1ecc6299db9ec823/failure-0.1.1/src/backtrace/internal.rs:44
   3:        0x10fd404fe - failure::backtrace::Backtrace::new::h3969a68773a2cd75
                        at /Users/parkr/.cargo/registry/src/github.com-1ecc6299db9ec823/failure-0.1.1/src/backtrace/mod.rs:98
   4:        0x10f802540 - failure::error::{{impl}}::from<std::io::error::Error>
                        at /Users/parkr/.cargo/registry/src/github.com-1ecc6299db9ec823/failure-0.1.1/src/error.rs:34
[...]

PR #84, opened 2018-02-01
Retry turning PID into process handle if it fails

jvns:
from @parkr's great suggestion on https://github.com/rbspy/rbspy/issues/80 -- previously we weren't ever retrying turning a PID into a task if it failed. This changes it so that we retry.

I forgot about this before because on Linux `.try_into_process_handle()` always succeeds, but on Mac it can fail and have races, so it makes sense to retry it.

### DIFF
--- a/src/initialize.rs
+++ b/src/initialize.rs
@@ -2,6 +2,7 @@ use copy::*;
 use copy;
 use failure::Error;
 use failure::ResultExt;
+use failure::Fail;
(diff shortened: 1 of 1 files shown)


luser:
FYI if you're spawning the process on Mac, you might want to use my [`spawn-task-port` crate](https://crates.io/crates/spawn-task-port). I don't know how well that works in the face of things like rbenv spawning child processes, though.

PR #86, opened 2018-02-01
Add subprocesses flag and flamegraph each

liaden:
🚧

- [x] Code
- [ ] Docs
- [ ] Docker tests

Pending things to be tackled in subsequent PRs:
- `--only-subprocesses`
- detecting new processes being forked
- merging the stacks together and generating one large flamegraph.

### Example of output:
```
Recording data to /home/jjohnson10/.cache/rbspy/records/rbspy-2018-02-01-OX2mHGKEqu.txt
Recording data to /home/jjohnson10/.cache/rbspy/records/rbspy-2018-02-01-oBvBMfq0nF.txt
Recording data to /home/jjohnson10/.cache/rbspy/records/rbspy-2018-02-01-H7QbbVLA46.txt
Recording data to /home/jjohnson10/.cache/rbspy/records/rbspy-2018-02-01-jRKjqdoglT.txt
^CInterrupted.
Writing flamegraph to /home/jjohnson10/.cache/rbspy/records/rbspy-2018-02-01-jRKjqdoglT.svg
Writing flamegraph to /home/jjohnson10/.cache/rbspy/records/rbspy-2018-02-01-H7QbbVLA46.svg
Writing flamegraph to /home/jjohnson10/.cache/rbspy/records/rbspy-2018-02-01-oBvBMfq0nF.svg
Writing flamegraph to /home/jjohnson10/.cache/rbspy/records/rbspy-2018-02-01-OX2mHGKEqu.svg
```


luser:
It doesn't have an API that's immediately usable for this purpose, but I wrote [a crate](https://github.com/luser/tracetree) to use ptrace to track all the child processes of a process. You could probably refactor it into something that rbspy could use. See the [`ProcessTree::spawn` method](https://github.com/luser/tracetree/blob/587eaaa90ad2469b37a6a8568e024276e99e11dc/src/lib.rs#L71) for all the interesting bits.

jvns:
@luser interesting thanks! so far we've avoided using using ptrace in rbspy because I don't think it's necessary to interrupt/pause the process to profile it but that's good to know about!

luser:
I agree that it's pretty invasive and I don't love it! Unfortunately AFAIK there's no other way to track child processes on Linux. I don't think the kernel even maintains this info on the parent side, I think it's purely stored as the `ppid` on the child process. Tools like `pstree` seem to iterate over all of `/proc/*` to find the info they need. (That may be fast enough to be feasible for rbspy!)

On Mac there is a dedicated API for getting just fork/exec/exit notifications, I wrote some notes on a GitHub issue but didn't get around to trying it out: https://github.com/luser/tracetree/issues/1

LLDB uses that API so I suspect it's the right thing to do!

liaden:
@luser My thought for future work was to use https://github.com/brendangregg/perf-tools/blob/master/execsnoop in the master process to be notified of a new child process and then spinning up a new thread to record that process. There would be some some amount of amount of latency between the `fork`ed process running and when rbspy would start monitoring though.

src/descendents.rs
```diff
@@ -0,0 +1,51 @@
...
+            Ok(proc_folder) => {
+                let pid_os_str = proc_folder.file_stem().expect("Glob pattern yields unexpected result");
+                let pid: pid_t = ::std::ffi::OsStr::to_str(pid_os_str).expect("huh").parse().expect("Glob pattern??");
+                let pid_status = ::procinfo::pid::status(pid).expect("Did not load pid status info");
+
+                pid_map.entry(pid_status.ppid).or_insert(vec!()).push(pid);
+            }
+        }
+    }
+    return pid_map;
```
jvns: i think it would work to write an integration test in here that tests the subprocess handling (which has the benefit of not relying on Docker, so the tests will work on Mac too)

src/main.rs
```diff
@@ -188,11 +192,60 @@ fn record(
...
+
+        let child_pid = std::thread::spawn(move || {
+            record(
+                format.outputter(),
+                &output_filename(&home_path, maybe_filename).expect("Error with output file"),
+                pid,
+                sample_rate,
+                maybe_duration,
+                done
+            ).expect("Failure in record");
```
jvns: would be good to have better error handling here -- in general it's good to minimize use of `expect`. Maybe we could send the error back to the main thread and then return it from `parallel_record`?

src/descendents.rs
```diff
@@ -0,0 +1,51 @@
...
+    let mut pid_map: HashMap<pid_t, Vec<pid_t>> = HashMap::new();
+    let proc_files = glob::glob("/proc/[0-9]*/").expect("Could not read glob pattern");
+
+    for proc_result in proc_files {
+        match proc_result {
+            Err(_) => (),
+            Ok(proc_folder) => {
+                let pid_os_str = proc_folder.file_stem().expect("Glob pattern yields unexpected result");
+                let pid: pid_t = ::std::ffi::OsStr::to_str(pid_os_str).expect("huh").parse().expect("Glob pattern??");
+                let pid_status = ::procinfo::pid::status(pid).expect("Did not load pid status info");
```
jvns: instead of having all these `.expect`s we could make `map_parents_to_children` return a `Result<HashMap<pid_t, Vec<pid_t>>` and turn them into errors.

ci/ruby-programs/ruby_forks.rb
```diff
@@ -0,0 +1,11 @@
...
+
+pid1 = fork do
+  pid1_1 = fork { loop { sleep SLEEP_TIME } }
+
+  loop { sleep SLEEP_TIME }
+end
+
+pid2 = fork { loop { sleep SLEEP_TIME } }
+
+loop { sleep SLEEP_TIME }
```
jvns: i like this test script!

jvns:
this is looking good! what are your thoughts about making it work on Mac? (which doesn't have `/proc`). one (probably bad) idea is to just run `pgrep -P $PID` as a subprocess on Mac. mostly suggest that because I am currently sorta grumpy about Mac systems APIs.

src/main.rs
liaden: From https://doc.rust-lang.org/1.17.0/std/thread/fn.spawn.html it seems like panicing might be acceptable in this case:
> If the child thread panics, join will return an Err containing the argument given to panic.

jvns:
I'll be happy to fix the merge conflicts here when this is ready to merge. I have some ideas for how to structure this in a way that will make it fit in better, will send you a PR shortly.

jvns:
Spent a good chunk of the day today working towards a way to merge this and sent you a couple of PRs (small: https://github.com/liaden/rbspy/pull/1, bigger: https://github.com/liaden/rbspy/pull/2).

jvns:
spent a while rebasing this today, so continuing in #109

PR #89, opened 2018-02-01
Don't require root to get the version

### LINKED ISSUE #88: `-V` shouldn't require root
Whilst putting together the Homebrew formula for this package (in Homebrew/homebrew-core#23575) I noticed that `rbspy -V` requires root permissions despite (on the surface) not needing these elevated permissions. Is this by design and I'm missing something here? Or should it not worry about needing root when calling this?

jvns:
Fixes #88 

rbspy doesn't work unless you're root on Mac. However we should wait to make that check until we're done parsing the command line arguments.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -106,6 +106,13 @@ fn do_main() -> Result<(), Error> {
 
     let args = Args::from_args()?;
 
+    if cfg!(target_os = "macos") {
+        let ok = sudo_if_not_root();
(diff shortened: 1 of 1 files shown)


jacobbednarz:
🎉 thanks a bunch! Can confirm this got it after rebuilding from `master`.

```
$ id -u
501

$ ./target/release/rbspy -V
rbspy
```

PR #69, opened 2018-01-26
Mac support

jvns:
Preliminary Mac support! In my testing so far, this works with Ruby (as long as it's installed by rbenv) on Mac.

This does **not** work with system Ruby on Mac. There are a couple issues with adding system Ruby support:

1. SIP -- getting a port to talk to system Ruby is impossible if you have SIP enabled on your Mac (tested this today). I believe this isn't fixable, the whole point of SIP is to prevent other programs from instrumenting your program in exactly this way.
2. my implementation of `get_process_maps` right now doesn't support dynamically loaded libraries, and all the symbols in system Ruby (at least on the High Sierra mac I'm working with) are in a dynamically loaded library which I haven't been about to find the address or path of yet. more notes about the difficulties there in [this blog post I wrote today](https://jvns.ca/blog/2018/01/26/mac-memory-maps/)

issue 2 there is definitely possible to work around but I think it might take me another 2 days so I think it's probably better to merge this as is. (and the SIP issue makes me wonder if it's even worth trying to support system Ruby, my guess is most people have SIP enabled)

Still need to add CI for Mac but wanted to get this out there.

PR #91, opened 2018-02-07
remove "Mac support" from Missing Features list

djbender:
It looks like mac support was added in #69 so I thought maybe this was missed in 5f7b3615. I thought Macs still weren't supported due to this.


jvns:
thank you!

PR #93, opened 2018-02-09
Report C functions in the Ruby stack as "unknown C function"

jvns:
Previously we were leaving C functions out of the stack that rbspy reports entirely. A most correct way to handle this is to instead report them as "unknown C function" -- this still doesn't give much insight into what the unknown C function **is** exactly but is a big improvement.

This PR both updates the code to report "unknown C function" when it think that's what happening, and adds a test with a core dump from a program which is running a C function. The new core dump would have previously returned an error from `get_stack_trace` and now returns an "unknown C function" stack.

This was causing an issue with profiling Ruby HTTP servers (for instance Sinatra), which sometimes spend a lot of time in the `select` syscall with no other Ruby function running -- rbspy would think 80% of the stacks were errors and give up.

### DIFF
--- a/src/ruby_version.rs
+++ b/src/ruby_version.rs
@@ -23,7 +23,7 @@ macro_rules! ruby_version_v_1_9_1(
             get_pos!(rb_iseq_struct);
             get_lineno_1_9_0!();
             get_stack_frame_1_9_1!();
-            is_stack_base_1_9_0!();
+            stack_field_1_9_0!();
         }
         ));
 
@@ -44,7 +44,7 @@ macro_rules! ruby_version_v_1_9_2_to_3(
             get_pos!(rb_iseq_struct);
             get_lineno_1_9_0!();
             get_stack_frame_1_9_2!();
-            is_stack_base_1_9_0!();
+            stack_field_1_9_0!();
         }
         ));
 
@@ -76,7 +76,7 @@ macro_rules! ruby_version_v_2_0_to_2_2(
             get_pos!(rb_iseq_struct);
             get_lineno_2_0_0!();
             get_stack_frame_2_0_0!();
-            is_stack_base_1_9_0!();
+            stack_field_1_9_0!();
         }
 ));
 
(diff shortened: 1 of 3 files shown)


src/ruby_version.rs
```diff
@@ -123,19 +136,29 @@ macro_rules! get_stack_trace(
...
+            if stack_field(&thread) as usize == 0 {
+                return Ok(vec!(unknown_c_function()));
+            }
             let mut trace = Vec::new();
             let cfps = get_cfps(thread.cfp as usize, stack_base(&thread) as usize, source)?;
             for cfp in cfps.iter() {
-                if cfp.iseq as usize == 0  || cfp.pc as usize == 0 {
-                    debug!("Either iseq or pc was 0, skipping CFP");
+                if cfp.iseq as usize == 0 {
+                    // this means this stack frame is a C function, note that and continue
```
kamalmarhubi: is this 100% true verified by reading mri source? or is it an assumption you're making?

jvns: the reason I believe this is that I grepped the Ruby source for `vm_push_frame` calls that set the `iseq` argument to 0 and they all have VM_FRAME_FLAG_CFRAME in them. I don't think I can check for that flag directly. 

```
vm.c:    vm_push_frame(th, 0 /* dummy iseq */, VM_FRAME_MAGIC_DUMMY | VM_ENV_FLAG_LOCAL | VM_FRAME_FLAG_FINISH | VM_FRAME_FLAG_CFRAME /* dummy frame */,
vm_eval.c:	vm_push_frame(th, 0, VM_FRAME_MAGIC_CFUNC | VM_FRAME_FLAG_CFRAME | VM_ENV_FLAG_LOCAL, recv,
vm_insnhelper.c:    vm_push_frame(th, 0, VM_FRAME_MAGIC_CFUNC | VM_FRAME_FLAG_CFRAME | VM_ENV_FLAG_LOCAL,
vm_insnhelper.c:    vm_push_frame(th, NULL, VM_FRAME_MAGIC_CFUNC | VM_FRAME_FLAG_CFRAME | VM_ENV_FLAG_LOCAL, recv,

```

PR #95, opened 2018-02-12
Add comments/docs for callgrind module

vasi:
Also new rustfmt.

Please read these over and validate that the comments are actually helpful. I'm happy to make changes to improve it!

### DIFF
--- a/src/callgrind.rs
+++ b/src/callgrind.rs
@@ -4,38 +4,62 @@ use std::io;
 
 use initialize::StackFrame;
 
+// Stats about the relationship between two functions, one of which
+// calls the other.
 #[derive(Debug)]
 struct Call {
+    // How many times has the outer function called the inner one?
     count: usize,
+
+    // How many samples were found inside the inner function
+    // (including all sub-functions), when called by the outer?
     inclusive: usize,
 }
 
+// Stats about a single function.
 #[derive(Debug, Default)]
 struct Location {
+    // How many samples were found inside this function only, not
+    // including calls to sub-functions?
     exclusive: usize,
+
+    // Data about the calls from this function to other functions.
     calls: HashMap<StackFrame, Call>,
 }
 
+// Stats about all functions found in our samples.
 #[derive(Default, Debug)]
 struct Locations(HashMap<StackFrame, Location>);
 
+// Information about a function currently on the stack.
 #[derive(Debug)]
 struct StackEntry {
     frame: StackFrame,
+
+    // How many samples were found inside this call only?
     exclusive: usize,
+
+    // How many samples were found in this call, and sub-calls?
     inclusive: usize,
 }
 
+// Tracks statistics about a program being sampled.
 #[derive(Default, Debug)]
 pub struct Stats {
-    // Stored with the root of the callgraph at the start.
+    // The current stack, along with tracking information.
+    // The root function is at element zero.
     stack: Vec<StackEntry>,
+
+    // Overall stats about this program.
     locations: Locations,
 }
 
 impl Locations {
+    // Get the current stats for a StackFrame. If it's never been seen before,
+    // automatically create an empty record and return that.
     fn location(&mut self, frame: &StackFrame) -> &mut Location {
         if !self.0.contains_key(frame) {
+            // Never seen this frame before, insert an empty record.
             let loc = Location {
                 ..Default::default()
             };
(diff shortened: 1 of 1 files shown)


src/callgrind.rs
```diff
@@ -4,38 +4,62 @@ use std::io;
...
 
+// Stats about the relationship between two functions, one of which
+// calls the other.
 #[derive(Debug)]
 struct Call {
+    // How many times has the outer function called the inner one?
     count: usize,
+
+    // How many samples were found inside the inner function
+    // (including all sub-functions), when called by the outer?
```
jvns: I don't understand the difference between the `count` and `inclusive` fields in this struct. Given 2 functions A and B, this is giving me information about calls from A to B. There are 2 possible ways I can think to measure calls from A to B given a bunch of stack traces:

1. count all direct calls A -> B
2. count all calls from A which eventually get to B somehow (possibly with multiplicity)

how do these 2 ways relate to the `count` and `inclusive` fields?

src/callgrind.rs
```diff
@@ -4,38 +4,62 @@ use std::io;
...
     inclusive: usize,
 }
 
+// Stats about a single function.
 #[derive(Debug, Default)]
 struct Location {
+    // How many samples were found inside this function only, not
+    // including calls to sub-functions?
     exclusive: usize,
```
jvns: does this `exclusive` field have the same value as the `exclusive` field on StackEntry for the corresponding function?

src/callgrind.rs
```diff
@@ -4,38 +4,62 @@ use std::io;
...
     inclusive: usize,
 }
 
+// Tracks statistics about a program being sampled.
 #[derive(Default, Debug)]
 pub struct Stats {
-    // Stored with the root of the callgraph at the start.
+    // The current stack, along with tracking information.
+    // The root function is at element zero.
     stack: Vec<StackEntry>,
```
jvns: is it true that we only store this stack to improve performance while calculating callgrind data?

src/callgrind.rs
```diff
@@ -44,82 +68,143 @@ impl Locations {
...
         while common < max_common && &self.stack[common].frame == rev[common] {
             common += 1;
         }
 
-        // Pop old items
+        // 2. Items only on the previous stack won't be kept, so we have to
+        // integrate them into our stats.
         while self.stack.len() > common {
+            // For each entry, pop it from our stored stack, and track its
+            // exclusive sample count.
```
jvns: I don't understand why **all** of the entries in the stack get their exclusive count updated. From the definition of exclusive above  I thought only the top thing on the stack (`E` in the diagram above) would get its exclusive count incremented, because we can't know that any time was spent in `C` or `D`

exclusive definition I read:

> How many samples were found inside this function only, not including calls to sub-functions?

src/callgrind.rs
jvns: I can't tell if this means the same thing as "self time" (number of times function appears at the top of the stack) or not.

jvns:
Thanks for writing this up! This helped me understand what's going on a bit more even though I have a lot of questions still. I'm going to tell you what I think the basic structure of this code is and you can tell me I'm right!

The Callgrind format encodes performance data basically as a graph, where the nodes are functions (like `a`) and the edges are statistics about calls between functions. The `Locations` struct is where that graph is stored, and to print out the callgrind file at the end we iterate over that struct in a pretty straightforward way.

Unlike the flamegraph format (which doesn't care about the order of the stack traces you've collected at all), the callgrind format **does** care about the order. The callgrind format kind of implicitly assumes that we have a tracing profile of our program and that you can get exact counts of the number of calls between every 2 functions. Since rbspy is a sampling profiler, this means we have to make some assumptions to make this format work!

**counting function calls**

How do we count function calls from a -> b in a sampling profiler? The main assumption we make is that if we have a stack (a,b,c) followed by another one with a common prefix (a,b,f,g), then that represents the same function call `a -> b`. This isn't necessarily true but if we're sampling at a high enough rate it's a reasonable assumption.

here's an example illustrating how we count function calls. let's assume we have these 4 stack traces.

```
a b c d d g x
a b c d d  // count  d -> g and  g -> x
a b c d // count  d -> d
a b e f g // count b -> c & c -> d
```

So if we take the function pair `b -> c` in this example, the Call struct would be `{count: 1, inclusive: 3}` -- there are 3 stack traces including the `b -> c` pair, but only 1 call from `b -> c`.

src/callgrind.rs
vasi: Yes, those are the same AIUI

src/callgrind.rs
jvns: I figured it eventually -- the magic happens down on line 174.

vasi:
Yes, that's exactly right! I definitely should have clarified the assumption about consecutive similar frames being "one function call"

src/callgrind.rs
vasi: Yeah, I should call out that there are two super-steps here:

A. Reconcile the previous and current stack (includes all of 1, 2, 3)
B. Actually count samples

jvns:
ok cool! i might actually comment this myself (with the parts that I don't think are obvious) and send you a PR with those comments?

vasi:
👍

jvns:
was going to send you a PR but I forgot. here's a commit with some more examples: https://github.com/rbspy/rbspy/commit/020f811cf1b278a40cf7f133c4045abcb4e793e1

vasi:
Looks great! 👍 

Nitpick: I should have caught this earlier, but this looks a bit off:

> the flamegraph format (which doesn't care about the order of the stack traces you've collected at all)

Flamegraphs definitely care about order, right? If you shuffled all your samples, it'd look pretty odd.

I think we're trying to say something about how flamegraphs don't care about the relationship between two consecutive traces? Though I'm not sure if that's true either—I think flamegraphs make the same assumption about consecutive similar traces being "the same". I'll think about this a bit more.

PR #101, opened 2018-02-15
Accentuated record by spawning process method in README

tvanderpol:
In my first two reads of the README I managed to miss the spawn process method,
it is possible others are similarly overlooking this great feature.

Please let me know if something should change for this PR to be a good fit!

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,19 +47,29 @@ sudo rbspy snapshot --pid $PID
 
 ### Record
 
-Record records stack traces from your process and generates a flamegraph. You can either give it the
-PID of an already-running process to record, or ask it to execute and record a new Ruby process.
+Record records stack traces from your process and generates a flamegraph.
 
 This is useful when you want to know what functions your program is spending most of its time in.
 
+You can record stack traces in two different ways, by PID or by executing a new ruby process.
+
+#### Record by PID
 ```
 sudo rbspy record --pid $PID
 # recording a subprocess doesn't require root access on Linux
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,19 +47,29 @@ sudo rbspy snapshot --pid $PID
...
 ```
 sudo rbspy record --pid $PID
 # recording a subprocess doesn't require root access on Linux
 # though on Mac rbspy always needs to run as root
+```
+
+#### Record by executing the process through rbspy
+```
 rbspy record ruby myprogram.rb
+# this will require sudo or root access on both mac and linux
```
tvanderpol: I am not sure this is true! I inferred from the comment above but please let me know if I'm wrong (no linux install to check atm)

README.md
```diff
@@ -47,19 +47,29 @@ sudo rbspy snapshot --pid $PID
...
 ```
 
-**Optional Arguments**
+#### Optional Arguments
+
+These work regardless of how you started the recording.
```
tvanderpol: I am similarly uncertain this is true, though it seems likely.

jvns:
Thanks so much for this! This definitely makes it clearer.

The root / not-root thing is a bit confusing -- the situation is that `rbspy record ruby my-process.rb` needs to run as root on Mac, but not Linux (because of how the underlying system calls rbspy uses works)

I added some explanations https://github.com/rbspy/rbspy/commit/dc8af2b87d67688519b87f5de911bf22fbf2de77 which I hope will make things clearer!

PR #102, opened 2018-02-15
Add a sketch of a versioned storage API

kamalmarhubi:
Versioning allows two things:
- decoupling inner data format from on disk format
- accepting older on-disk formats in newer rbspy versions, eg, to allow
  new visualisations on old `rbspy record` outputs


kamalmarhubi:
Evident: I still haven't figured out how to use failure properly.

jvns:
this code is merged now!!

PR #108, opened 2018-02-21
Add stack trace struct containing PID

jvns:
This updates the data storage format for rbspy to include a PID for each stack trace. The idea here is that sometimes we'll sample stack traces from multiple processes, so it's good to include a PID for each stack trace. This is to support the subprocess profiling work by @liaden.

### DIFF
--- a/src/core/initialize.rs
+++ b/src/core/initialize.rs
@@ -28,13 +28,13 @@ use std;
  *   * Find the right stack trace function for the Ruby version we found
  *   * Package all that up into a struct that the user can use to get stack traces.
  */
-pub fn initialize(pid: pid_t) -> Result<StackTraceGetter, Error> {
+pub fn initialize(pid: pid_t) -> Result<StackTraceGetter<ProcessHandle>, Error> {
     let version = get_ruby_version_retry(pid).context("Couldn't determine Ruby version")?;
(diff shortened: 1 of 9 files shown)


src/core/initialize.rs
```diff
@@ -76,12 +76,24 @@ impl PartialOrd for StackFrame {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize)]
+pub struct StackTrace {
+    pub trace: Vec<StackFrame>,
+    pub pid: pid_t,
```
kamalmarhubi: should this be an option? how will the various reporty things handle 0?

(at least deserves comment I think)

src/storage/mod.rs
```diff
@@ -113,7 +107,7 @@ fn read_version(r: &mut Read) -> Result<Version, StorageError> {
     }
 }
 
-pub(crate) fn from_reader<R: Read>(r: R) -> Result<Vec<Vec<StackFrame>>, Error> {
+pub(crate) fn from_reader<R: Read>(r: R) -> Result<v1::Data, Error> {
```
kamalmarhubi: you need another match arm in here for v1 :-)

src/core/initialize.rs
jvns: Made it an option!

PR #109, opened 2018-02-22
Add --subprocesses flag to rbspy record

jvns:
This is #86, but rebased on master and squashed into a single commit. I also removed the glob & procinfo dependencies in favour of just parsing /proc/PID/status manually (since it's like 10 lines of code to parse that file and I'd sorta like to keep dependencies under control).

planning to write maybe marginally better tests then merge it!

### DIFF
--- a/ci/ruby-programs/ruby_forks.rb
+++ b/ci/ruby-programs/ruby_forks.rb
@@ -1,0 +1,11 @@
+SLEEP_TIME = 2
+
+pid1 = fork do
+  pid1_1 = fork { sleep SLEEP_TIME }
+
+  sleep SLEEP_TIME
+end
+
+pid2 = fork { sleep SLEEP_TIME }
+
+sleep SLEEP_TIME
--- a/src/main.rs
+++ b/src/main.rs
@@ -44,22 +44,26 @@ use libc::pid_t;
 use failure::Error;
 use failure::ResultExt;
 
+use std::collections::HashSet;
 use std::fs::{DirBuilder, File};
 use std::path::{Path, PathBuf};
 use std::env;
 use std::process::Command;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
-use std::time::Instant;
+use std::time::{Instant, Duration};
 use std::os::unix::prelude::*;
+use std::sync::mpsc::{sync_channel, channel, SyncSender, Receiver};
 
 pub mod core;
 pub mod ui;
 pub(crate) mod storage;
 
 use core::initialize::initialize;
+use core::types::StackTrace;
 use core::copy::MemoryCopyError;
 use ui::output;
+use ui::descendents::descendents_of;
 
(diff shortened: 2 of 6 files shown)


src/main.rs
```diff
@@ -488,6 +564,10 @@ fn arg_parser() -> App<'static, 'static> {
                     ).conflicts_with("cmd")
                         .required(false),
                 )
+                .arg(
+                    Arg::from_usage( "-s --subprocesses='Record all subprocesses of the given PID or command'")
+                        .required(false)
```
liaden: `conflicts_with("cmd")`?

src/ui/descendents.rs
```diff
@@ -0,0 +1,107 @@
...
+fn test_get_descendents_depth_2() {
+    let mut map = HashMap::new();
+    map.insert(1, vec![2, 3]);
+    map.insert(2, vec![4]);
+    let desc = get_descendents(1, map).unwrap();
+    assert_eq!(desc, vec![1, 3, 2, 4]);
+}
+
+// parses /proc/<pid>/status format
+fn status_file_ppid(status: &str) -> Result<pid_t, Error> {
```
liaden: Was the other dependency to get the proc_info too slow or memory extensive?

jvns: nope! this is really a personal preference thing -- like right now I feel like it's better to depend on a very small amount of code that's easy to understand instead of a large amount of code in a crate I don't understand. I'm not really sure if this is better or worse honestly.

jvns:
pushed another commit that lets it not conflict with `cmd`! Basically this spawns a thread that continually monitors for new child processes and starts recording them.

I also added a simple test for the process-spawning logic.

ci/ruby-programs/ruby_forks.rb
```diff
@@ -1,11 +1,11 @@
 SLEEP_TIME = 2
 
 pid1 = fork do
-  pid1_1 = fork { loop { sleep SLEEP_TIME } }
```
liaden: Definitely good to remove the `loop`s.

src/main.rs
```diff
@@ -262,42 +268,178 @@ impl SampleTime {
...
+    Ok((trace_receiver, result_receiver))
+}
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_spawn_record_children_subprocesses() {
+    // We only run this test on linux because the subprocess code doesn't work on Mac yet
+    let mut process = std::process::Command::new("/usr/bin/ruby").arg("ci/ruby-programs/ruby_forks.rb").spawn().unwrap();
+    let pid = process.id() as pid_t;
+    let (_trace_receiver, result_receiver) = spawn_recorder_children(pid, true, 50, None).unwrap();
```
liaden: `test test_spawn_record_children_subprocesses has been running for over 60 seconds`

This test isn't terminating on travis. Given we are passing `None` for `maybe_stop_time`, I assume we are not able to detect that the subcommand has exited and thus we are still blocking on the child thread that is error on collecting stack traces?

PR #119, opened 2018-02-25
[Don't merge] Add a docker test that should work, but fails

jmeagher:
This is mostly a call for help in either figuring out what is wrong with the docker setup I'm using to run this example or help with finding what's missing in the Rust code. The errors this shows matches what I have seen trying to test on live systems running as docker containers. After adding this new test in ci it seems to work on the travis-ci servers. For me this is failing on a Mac with docker 17.12.0-ce and go go1.9.2.

Everything is running as root so that shouldn't be the cause of the permission error. It appears to be finding the right binary and symbol address. The attempt to copy the ruby version out of the running process is failing with: `DEBUG:rbspy::core::copy: permission_denied_error: details: Os { code: 1, kind: PermissionDenied, message: "Operation not permitted" }`. 

On this branch run: `docker build -t rbspy -f ci/Dockerfile.localtest . && docker run rbspy`

This fails with the following errors (with the few extra debug messages):

```
+ RUST_LOG=debug
+ ./rbspy record /usr/local/bin/ruby /short_program.rb
Press Ctrl+C to stop
DEBUG:rbspy::core::address_finder::os_impl: Found path: Some("/usr/local/bin/ruby")
DEBUG:rbspy::core::address_finder::os_impl: Found path: Some("/usr/local/lib/libruby.so.2.3.0")
DEBUG:rbspy::core::address_finder::os_impl: elf_symbol_value: symbol: Symbol: Value: 0x002422b8 Size: 0x0006 Type: data object Bind: global Vis: default Section: 13 Name: ruby_version
DEBUG:rbspy::core::address_finder::os_impl: get_symbol_addr: addr: 2422b8 range_start: 7ff592bf5000 load header: Program Header: Type: LOAD Offset: 0x00000000 VirtAddr: 0x00000000 PhysAddr: 0x00000000 FileSize: 0x2e1b24 MemSize: 0x2e1b24 Flags: R E Align: 0x200000
DEBUG:rbspy::core::initialize: pub get_ruby_version: address: 7ff592e372b8 process 8
DEBUG:rbspy::core::copy: copy_address_raw: addr: 7ff592e372b8
DEBUG:rbspy::core::copy: permission_denied_error: details: Os { code: 1, kind: PermissionDenied, message: "Operation not permitted" }
Wrote raw data to /root/.cache/rbspy/records/rbspy-2018-02-25-RlTxzUcfee.raw.gz
[...]

PR #121, opened 2018-02-26
Include sys_ptrace capability information in the error

jmeagher:
Add a more helpful message when permission denied errors arise. See #119 for a little background.

Note: This is a Linux specific error that came up for me under docker. Fix it in docker with `docker run --cap-add SYS_PTRACE`


jvns:
thanks, this is a better error message!

PR #123, opened 2018-03-17
Get stacks starting from ruby_current_vm instead of ruby_current_thread

jvns:
Previously we were getting stacks starting from `ruby_current_thread`. `(ruby_current_thread->cfp->...)`. This changes it so that instead start at the `ruby_current_vm` symbol, and do `ruby_current_vm->running_thread->cfp->...`.

The main thing this will get us (in my next PR, tomorrow or so) is easier access to whether the Ruby VM is currently garbage collecting or not, which is `ruby_current_vm->objspace->during_gc`.

### DIFF
--- a/src/core/address_finder.rs
+++ b/src/core/address_finder.rs
@@ -23,7 +23,7 @@ pub enum AddressFinderError {
 #[cfg(target_os = "macos")]
 mod os_impl {
     use crate::core::address_finder::AddressFinderError;
-    use crate::core::initialize::IsMaybeThreadFn;
+    use crate::core::initialize::IsMaybeVmFn;
     use proc_maps::{get_process_maps, MapRange};
     use proc_maps::mac_maps::{get_symbols, Symbol};
 
@@ -36,17 +36,12 @@ mod os_impl {
         proginfo.symbol_addr("_ruby_version")
     }
 
-    pub fn current_thread_address(
+    pub fn current_vm_address(
         pid: pid_t,
-        version: &str,
-        _is_maybe_thread: IsMaybeThreadFn,
+        _is_maybe_vm: IsMaybeVmFn,
     ) -> Result<usize, Error> {
         let proginfo = &get_program_info(pid)?;
-        if version >= "2.5.0" {
-            proginfo.symbol_addr("_ruby_current_execution_context_ptr")
-        } else {
-            proginfo.symbol_addr("_ruby_current_thread")
-        }
+        proginfo.symbol_addr("_ruby_current_vm")
     }
 
     struct Binary {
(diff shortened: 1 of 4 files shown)


src/core/address_finder.rs
```diff
@@ -152,19 +147,18 @@ mod os_impl {
...
-        version: &str,
-        is_maybe_thread: Box<Fn(usize, usize, ProcessHandle, &Vec<MapRange>) -> bool>,
+        is_maybe_vm: Box<Fn(usize, ProcessHandle) -> bool>,
     ) -> Result<usize, Error> {
         let proginfo = &get_program_info(pid)?;
-        match current_thread_address_symbol_table(proginfo, version) {
+        match current_vm_address_symbol_table(proginfo) {
             Some(addr) => Ok(addr),
             None => {
                 debug!("Trying to find address location another way");
```
kamalmarhubi: do you log anywhere that you were trying to find it in the symbol table? if not "another way" is kind of vague

src/core/address_finder.rs
```diff
@@ -255,34 +249,23 @@ mod os_impl {
 
         let i = slice
             .iter().enumerate()
-            .position({ |(i, &x)| is_maybe_thread(x, (i as usize) * (std::mem::size_of::<usize>() as usize) + read_addr, source, &proginfo.all_maps) })
+            .position({ |(i, &_x)| is_maybe_vm((i as usize) * (std::mem::size_of::<usize>() as usize) + read_addr, source) })
```
kamalmarhubi: style: you should be able to use `_` instead of `&_x` if you don't use it

src/core/ruby_version.rs
```diff
@@ -119,25 +119,45 @@ macro_rules! ruby_version_v2_5_x(
...
-            ruby_current_thread_address_location: usize,
+            ruby_current_vm_address_location: usize,
             process: &Process<T>,
             ) -> Result<StackTrace, MemoryCopyError> where T: CopyAddress {
             let source = &process.source;
-            let current_thread_addr: usize =
-                copy_struct(ruby_current_thread_address_location, source)?;
-            let thread: $thread_type = copy_struct(current_thread_addr, source)?;
+            let current_vm_addr: usize = copy_struct(ruby_current_vm_address_location, source)?;
+            let vm: rb_vm_t = copy_struct(current_vm_addr, source)?;
```
kamalmarhubi: any idea if this extra indirection has performance impact?

neosilky:
@jvns Just in the process of rebasing your PR onto master. I am just fixing tests :+1:

acj:
Closing since a lot has changed in the meantime and these changes would be hard to reconcile. We have code for getting the VM pointer now (part of support for ruby 3.x), and I'd still like to pull in garbage collection info if we can.

PR #127, opened 2018-03-23
Measure how much time is spent on CPU

jvns:
This measures how much time overall the program we're profiling is spending on the CPU! This opens up some exciting visualization possibilities -- now that we're recording whether or not each stack trace is on the CPU or not, we can do things like add an `--on-cpu` / `--off-cpu` to `rbspy report` to only include stack traces where the program was / wasn't using the CPU.

Only works on Linux so far.

### DIFF
--- a/src/core/ruby_version.rs
+++ b/src/core/ruby_version.rs
@@ -150,21 +150,50 @@ macro_rules! get_stack_trace(
 
         use crate::core::types::*;
         use crate::core::types::StackFrame;
+        use failure::Error;
+        use libc::pid_t;
 
         pub fn get_stack_trace<T>(
             ruby_current_thread_address_location: usize,
             process: &Process<T>,
             ) -> Result<StackTrace, MemoryCopyError> where T: CopyAddress {
+            let trace = get_stack_frames(ruby_current_thread_address_location, process)?;
+            let pid = process.pid;
+            let on_cpu = get_cpu(pid).ok();
+            Ok(StackTrace{on_cpu, trace, pid, thread_id: None})
+        }
+
+        fn get_cpu(maybe_pid: Option<pid_t>) -> Result<bool, Error> {
+            match maybe_pid {
+                None => Err(format_err!("no pid given")),
+                Some(pid) => on_cpu(pid),
+            }
+        }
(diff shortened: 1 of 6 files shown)


petethepig:
I looked into this a little bit, on macOS it's possible to use `task_info` call to get `pti_numrunning`, e.g:
```
#[cfg(target_os = "macos")]
fn on_cpu(pid: pid_t) -> Result<bool, Error>{
    match pidinfo::<TaskInfo>(pid, 0) {
        Ok(info) => Ok(info.pti_numrunning > 0),
        Err(err) => Ok(true),
    }
}
```

src/core/ruby_version.rs
```diff
@@ -150,21 +150,50 @@ macro_rules! get_stack_trace(
...
+                Some(pid) => on_cpu(pid),
+            }
+        }
+
+        #[cfg(target_os = "linux")]
+        fn on_cpu(pid: pid_t) -> Result<bool, Error> {
+            use std::fs::File;
+            use std::io::Read;
+            let mut contents = String::new();
+            File::open(format!("/proc/{}/status", pid))?.read_to_string(&mut contents)?;
```
petethepig, line 178: I found that this approach only works for the main thread of the app. So if you start a ruby process, start a thread and do things in that other thread `/proc/pid/status` won't reflect that.

Something like this works well though:
```
    #[cfg(target_os = "linux")]
    fn on_cpu(&self) -> Result<bool, Error> {
        for thread in self.process.threads()?.iter() {
            let active = thread.active()?;
            if active {
                return Ok(true);
            }
        }
        Ok(false)
    }
```

jvns: interesting -- it looks like under the hood remoteprocess is using `/proc/PID/task` to get information about the thread. Also it looks like remoteprocess supports this `threads()` function on non-Linux platforms which is great.

I'm not really working on rbspy features right now but I'd be happy to merge a version of this PR if you're interested in either improving this one or making a different better one. I think this would be great to have.

I'm also not totally sure about the performance impact of this PR. -- if it slows things down a lot then we should make sure to document that.

Issue #128, opened 2018-03-27
Get rid of "variant is never constructed: `MacPermissionDenied`" warning


jvns:
It's confusing for people trying to develop rbspy.

tvanderpol:
This is me dipping my toe into the rust waters so I could well have misunderstood the intent behind this issue!

PR #133, opened 2018-04-03
Removed MacPermissionDenied variant

tvanderpol:
It was causing confusion, per https://github.com/rbspy/rbspy/issues/128

### DIFF
--- a/src/core/address_finder.rs
+++ b/src/core/address_finder.rs
@@ -14,8 +14,6 @@ pub enum AddressFinderError {
     #[fail(display = "No process with PID: {}", _0)] NoSuchProcess(pid_t),
     #[fail(display = "Permission denied when reading from process {}. Try again with sudo and check sys_ptrace capability?", _0)]
     PermissionDenied(pid_t),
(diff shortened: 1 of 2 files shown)


jvns:
Hi! I realize now that I specified that issue kind of poorly (still learning how to make good first issues...), so here is a more clear explanation!

Basically the reason for that `MacPermissionDenied` error to be special cased (instead of just using `task_for_pid(...)`?) is that it lets us exit more quickly in `initialize.rs` -- normally it'll retry 50 times if something fails, but if there's a "permission denied" error we want to exit immediately because it'll never succeed.

The code actually compiles without warnings on Mac, because the `MacPermissionDenied` error code actually *is* used, it's just that it's only used on Mac and so you get compiler warnings on Linux.

So! I think we could fix this most easily (though it's maybe a bit gross) by adding this compiler macro above the 2 places where MacPermissionDenied is referenced saying to only compile the code declaring that error on Mac (there are 3, but one of them already has that flag)

```
    #[cfg(target_os = "macos")]
```

let me know if that makes sense!

jvns:
closing in favour of https://github.com/rbspy/rbspy/pull/147

PR #143, opened 2018-04-23
Add Ruby 2.2.10 support

tcaddy:
Add support for Ruby 2.2.10.

(I looked at previous pull requests to see which files to edit.)


jvns:
thanks very much for this!

Merged this manually in https://github.com/rbspy/rbspy/commit/4b0d7bb13dfc1a1a4673637fbeb2579d59b60d7a 

The `ruby_2_2_10.rs` file is supposed to be created by this script called `bindgen.sh` from the Ruby source code for 2.2.10, so I ran it and replaced the bindings file from this PR with the correct bindings file. That script only works on my laptop right now but at some point we should fix it so that other people can run it as well.

liaden:
@jvns It would be good to get it working for others because I have a fork of 1.9.3 for our legacy application that I am probably going to need to run bindgen and use that version of rbspy with the legacy application.

jvns:
@liaden oh interesting!! I'll put it on my list of things to do (or if you want to try to fix the script yourself that would be amazing)

PR #154, opened 2018-05-19
[Closes #148] Support --subprocesses flag on Mac

### LINKED ISSUE #148: Support --subprocesses flag on Mac
Right now this isn't supported because I haven't figured out how to get all the child process IDs of a process on Mac. 

All that would be required to implement this would be to implement a Mac version of this `get_proc_children` function which is meant to return all pairs of `<pid, parent pid>` on the system.

One possible way to do this (and maybe the easiest way) would be to run `ps` with the appropriate arguments and parse the output. this seems to be what tmm1's [implementation of pstree does](https://github.com/tmm1/pstree/blob/master/pstree.c)

https://github.com/rbspy/rbspy/blob/568ccd1be5ce2a433f2d1114ff3fe8746ce0c46e/src/ui/descendents.rs#L86-L102

akhramov:
Currently --subprocesses flag is not available on darwin platform,
because subprocess retrieving logic is Linux-specific.

This change
* Bumps `libproc` version to "0.3.1", because the new version
  introduces `proc_pid::pidinfo` function, useful for retrieving
  PPIDs.
* Implements darwin-specific `ui::descendents::descendents_of` based
  on the new `libproc` functionality.
* Enables `test_spawn_record_children_subprocesses` test for darwin
  and tweaks it not to use system Ruby for reasons described in #69

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -245,7 +245,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "libproc"
-version = "0.2.0"
+version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "errno 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


akhramov:
Please hold on with this request. I found a bug in libproc library, but the fix is on its way (andrewdavidmackenzie/libproc-rs/pull/8).

jvns:
thank you so much for this! so sorry for the delay in merging it -- I saw the "please hold on" comment but missed the update until now :)

akhramov:
It's okay. After all, I'm very happy to contribute to project that helped me a lot. Thank you!

PR #165, opened 2018-07-31
Add Windows Support

benfred:
Thanks for this project! I’ve found the code and blog posts on developing rbspy incredibly useful and inspiring.

This PR attempts to get rbspy working on Windows. It doesn’t totally work just yet, but I thought I would share where I’m at anyways.

In particular this PR:

* makes rbspy build on windows
* adds code to read virtual memory maps on windows and load symbols from windows processes
* Provides windows implementations of the address_finder::get_ruby_version_address and current_thread_address functions
* refactors proc_maps/mac_maps/win_maps into a subcrate, and has all OS implementations expose a common interface (get_process_maps/MapRange)

With these changes rbspy can connect up to a process on windows, and correctly determine the version of ruby running. 

Unfortunately when trying to copy the stack traces it fails with an error like:

	Copy error: stack base and cfp address out of sync. stack base: 28ce010, cfp address: 294df20

I'm a little stuck on exactly what is going wrong here to be honest. I was hoping someone here might have any ideas of where I'm going wrong here =)

Also, I'm interested in using the proc-maps code in another project I'm working on - and I was wondering if we could extract to a standalone crate? I think moving the proc-maps folder to its own repo and then updating the Cargo.toml file here should be sufficient. I'm willing to set this up, but wanted to make sure that this is something you'd be interested in first. This can be done independently of the rest of the windows changes here, since you probably don't merge this just yet.


jvns:
hi @benfred!! This is so exciting! I'd love to have Windows support and I'm really impressed by the progress here so far.

on the crate -- I'm not that excited about extracting proc-maps into a crate inside `rbspy` because I don't want to maintain a new crate. If you're interested in taking the code and creating a new crate that you maintain of course that would be fine :)

Can we start by just making this new `proc_maps` abstraction a module and not its own crate?

re how to fix this bug -- where I'd start would be to poke around at a Ruby process in Windows in gdb and make sure that what you're seeing with rbspy matches what you're seeing with gdb. Here's how to poke at it with gdb: https://jvns.ca/blog/2016/06/12/a-weird-system-call-process-vm-readv/.

That's how I developed rbspy initially -- just ran a single Ruby program in an infinite loop (this one: https://github.com/rbspy/rbspy/blob/master/ci/ruby-programs/infinite.rb), added a lot of print statements, and made sure that the addresses I was getting with rbspy always matched up with the addresses I got from gdb. Does that help?

jvns:
also -- would you be up for submitting a PR with the refactor of the proc_maps code into the new version with a better interface separately? I think that's a really reasonable change and that it would be easier to review if it was split out of this.

benfred:
Thanks for the suggestions! I'll let you know how it goes, I'm pretty new to Rust and haven't done much Windows programming recently - so it might take me a bit =)

I've created a new PR with just the linux/osx proc_maps changes into a submodule here: https://github.com/rbspy/rbspy/pull/166 .

I’ve left the windows proc-maps code out of that PR. Unfortunately I had to expose some extra functionality in the winapi-rs crate - [and my PR is waiting to be merged](https://github.com/retep998/winapi-rs/pull/653). It seems like just linking to a git repo in Cargo.toml like I am doing here means that you can’t upload to crates.io =(.

Also, totally unrelated to this - I noticed that rbspy leaks memory on OSX (uses about 1.4GB of memory a minute on OSX for me).  I fixed here https://github.com/luser/read-process-memory/pull/5 , and with that change rbspy memory usage stays constant at ~3mb.

benfred:
So - I spent a couple this afternoon looking into this, and I got rbspy working on windows!

screen shot 2018-08-04 at 8 11 57 pm

The problem was in the automatically generated ruby-structs from bindgen - the VALUE type was being defined as a std::os::raw::c_ulong  (because of this code https://github.com/ruby/ruby/blob/trunk/include/ruby/ruby.h#L94), which on windows is only 32 bits. For windows,  that ruby header checks the size of the ulong and typedefs VALUE to a 64-bit c_ulonglong.

I've updated this PR with the bindgen changes needed (and removed the previous attempt). 

It's still not ready to merge yet - but it at least works now =)

jvns:
fantastic! I'm away from a computer this week but will look on Saturday

kamalmarhubi:
@benfred :clap this is really amazing! I haven't looked in detail yet, but I'm inclined to merge this pretty soon and figure out what to do with `ruby-structs` afterwards. Especially since you already added appveyor CI.

src/main.rs
```diff
@@ -698,9 +719,14 @@ impl Args {
...
             ("record", Some(submatches)) => {
                 let format = value_t!(submatches, "format", OutputFormat).unwrap();
-                let raw_path = output_filename(&std::env::var("HOME")?, submatches.value_of("raw-file"), "raw.gz")?;
-                let out_path =
-                    output_filename(&std::env::var("HOME")?, submatches.value_of("file"), &format.extension())?;
+
+                #[cfg(unix)]
+                let home = &std::env::var("HOME")?;
+                #[cfg(windows)]
+                let home = &std::env::var("userprofile")?;
```
kamalmarhubi: filed #167 to clean this up later via either dirs or directories crate

src/ui/flamegraph.rs
```diff
@@ -82,6 +82,7 @@ mod tests {
     }
 }
 
+#[cfg(unix)]
```
kamalmarhubi: can you add a comment saying why this is cfg'ed out? also maybe `#[cfg(not(windows))]` depending? Is this because of the lack of perl?

src/ui/descendents.rs
```diff
@@ -121,3 +124,11 @@ fn get_proc_children() -> Result<Vec<(pid_t, pid_t)>, Error> {
 
     Ok(pids.iter().map(|&pid| pid as pid_t).zip(ppids).collect())
 }
+
+#[cfg(windows)]
+fn get_proc_children() -> Result<Vec<(pid_t, pid_t)>, Error> {
+    // TODO: this isn't currently implemented, but can be done
+    // using the tlhelp32 Process32First/Process32Next api from the
+    // winapi-rs crate: https://github.com/retep998/winapi-rs/blob/0.3/src/um/tlhelp32.rs#L94
+    Err(format_err!("--subprocesses option is not yet supported on windows"))
```
kamalmarhubi: could you change this to `unimplemented!(<msg here>)`, and instead do one of
-  `cfg` out the code that adds the option to the CLI (lhttps://github.com/rbspy/rbspy/blob/111689fe13f843ac6786855ac10ce66c3ad84ab7/src/main.rs#L655)
- fail when matching the command: lhttps://github.com/rbspy/rbspy/blob/111689fe13f843ac6786855ac10ce66c3ad84ab7/src/main.rs#L655

I think it'll be easier to do the second because the args are added in one huuuuuge expression.

src/main.rs
```diff
@@ -38,7 +39,10 @@ extern crate term_size;
 
 use chrono::prelude::*;
 use clap::{App, AppSettings, Arg, ArgMatches, SubCommand};
+#[cfg(unix)]
 use libc::pid_t;
+#[cfg(windows)]
+type pid_t = u32;
```
kamalmarhubi: are you sure this is the right size?

src/ui/descendents.rs
```diff
@@ -4,7 +4,10 @@ use std::fs::File;
 use std::io::Read;
 
 use failure::Error;
+#[cfg(unix)]
 use libc::pid_t;
+#[cfg(windows)]
+type pid_t = u32;
```
kamalmarhubi: could you add a `types` module at the top level that handles this cfg to avoid duplicating it?

scripts/bindgen.sh
```diff
@@ -41,3 +41,6 @@ echo "#![allow(non_upper_case_globals)]" > $OUT
 echo "#![allow(non_camel_case_types)]" >> $OUT
 echo "#![allow(non_snake_case)]" >> $OUT
 cat /tmp/bindings.rs >> $OUT
+
+# fix up generated bindings so that they compile/work on windows
+perl -pi -e "s/::std::os::raw::c_ulong;/u64;/g" $OUT
```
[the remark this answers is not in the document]
benfred: Good catch! I changed to a usize here: https://github.com/rbspy/rbspy/pull/165/commits/9364de325bb778233cdace5b173c0a1d431f84d2 (I replaced the original commit with this one to hopefully keep this PR easier to review).

For the commit message, I meant both windows and os/linux - I've updated so it's hopefully clearer.

src/ui/flamegraph.rs
benfred: It's a combination of lack of perl, and that the tempdir.close() function fails with a 'directory is not empty' error on windows. I've updated the comment to track this.

src/ui/descendents.rs
benfred: I went with the second option on this, since the first one would have been a pretty messy change.

I also made the subprocesses option 'hidden' on windows which might help a bit

src/main.rs
benfred: I *think* it’s the right size.  Inside the win32 api, pid’s are a DWORD type - which are unsigned long’s (which are 32 bits on windows): https://github.com/retep998/winapi-rs/blob/0.3/src/shared/minwindef.rs#L21 .

src/ui/descendents.rs
benfred: Done - but I used the core::types module instead of creating a new one

benfred:
@kamalmarhubi Thanks for the feedback! I think I addressed most of the issues you found in the last commit

src/main.rs
```diff
@@ -18,6 +18,7 @@ extern crate libproc;
 #[cfg(target_os = "macos")]
 extern crate mach;
 extern crate nix;
+extern crate proc_maps;
```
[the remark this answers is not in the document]
benfred: Added a comment to that PR https://github.com/rbspy/rbspy/pull/166#issuecomment-412156063

jvns:
This is incredible! Thank you so much. I've:

* cut a new release (0.2.9 -> 0.3.0) since this is a huge improvement
* added you as a collaborator on the project
* given AppVeyor access to the rbspy repository so that we always test building on Windows (and checked that it works: https://github.com/rbspy/rbspy/pull/168)
* updated the README to say that rbspy supports Windows now

PR #166, opened 2018-08-04
refactor proc_maps into a submodule

benfred:
This code pulls out proc_maps/mac_maps into a submodule - and attempts to have
them present a unified interface.

Each OS version exposes a MapRange struct and get_process_maps function. The
get_process_maps function has the same signature for both linux and OSX, and
the MapRange struct exposes the same getter methods, with the idea that you
can use them identically across operating systems.

### DIFF
--- a/src/core/proc_maps/linux_maps.rs
+++ b/src/core/proc_maps/linux_maps.rs
@@ -1,32 +1,33 @@
+use libc;
 use std;
 use std::fs::File;
 use std::io::Read;
-use libc::pid_t;
+
+pub type Pid = libc::pid_t;
+
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct MapRange {
-    pub range_start: usize,
-    pub range_end: usize,
-    pub offset: usize,
-    pub dev: String,
-    pub flags: String,
-    pub inode: usize,
-    pub pathname: Option<String>,
+    range_start: usize,
+    range_end: usize,
+    offset: usize,
+    dev: String,
+    flags: String,
+    inode: usize,
+    pathname: Option<String>,
 }
 
 impl MapRange {
-    pub fn contains_addr(&self, addr: usize) -> bool {
-        addr >= self.range_start && addr <= self.range_end
-    }
+    pub fn size(&self) -> usize { self.range_end - self.range_start }
+    pub fn start(&self) -> usize { self.range_start }
+    pub fn filename(&self) -> &Option<String> { &self.pathname }
+    pub fn is_exec(&self) -> bool { &self.flags[2..3] == "x" }
+    pub fn is_write(&self) -> bool { &self.flags[1..2] == "w" }
+    pub fn is_read(&self) -> bool { &self.flags[0..1] == "r" }
 }
 
-pub fn maps_contain_addr(addr: usize, maps: &Vec<MapRange>) -> bool {
-    maps.iter().any({ |map| map.contains_addr(addr) })
-}
-
-pub fn get_proc_maps(pid: pid_t) -> Result<Vec<MapRange>, std::io::Error> {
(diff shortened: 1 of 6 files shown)


src/core/proc_maps/linux_maps.rs
```diff
@@ -1,32 +1,33 @@
...
-}
-
-pub fn maps_contain_addr(addr: usize, maps: &Vec<MapRange>) -> bool {
-    maps.iter().any({ |map| map.contains_addr(addr) })
+    pub fn size(&self) -> usize { self.range_end - self.range_start }
+    pub fn start(&self) -> usize { self.range_start }
+    pub fn filename(&self) -> &Option<String> { &self.pathname }
+    pub fn is_exec(&self) -> bool { &self.flags[2..3] == "x" }
+    pub fn is_write(&self) -> bool { &self.flags[1..2] == "w" }
+    pub fn is_read(&self) -> bool { &self.flags[0..1] == "r" }
```
jvns: it doesn't really matter but it might be nice (since there's sort of an implicit interface here) to implement a real interface.

src/core/proc_maps/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+pub use self::linux_maps::{get_process_maps, MapRange, Pid};
+
+fn map_contain_addr(map: &MapRange, addr: usize) -> bool {
+    let start = map.start();
+    (addr > start) && (addr < (start + map.size()))
+}
+
+pub fn maps_contain_addr(addr: usize, maps: &[MapRange]) -> bool {
+    maps.iter().any({ |map| map_contain_addr(map, addr) })
+}
```
jvns: this is much nicer

jvns: I find this a bit confusing because the `maps_contain_addr` code is actually only used on Linux, and so it seems misleading today to write a generic version of it. 

here's a commit that might clarify things -- it takes all the usages of `maps_contain_addr` and marks them to only compile on Linux. I think that code will still compile on Mac. https://github.com/rbspy/rbspy/commit/e3f52831f369f48b30fa7976b0a0c6b69021dfbe

benfred: I agree with you that having this generic functionality isn’t needed yet - but the cool thing is that that the ‘could_be_thread’ and ’is_maybe_thread’ functions will work on OSX without any further changes like this. 

Like, we could support stripped binaries on OSX pretty easily here - just by extracting the __bss section address from the mach-o ruby binaries and then using the same code as on linux for scanning for the currentthread addresses.

Your change doesn't quite compile on OSX, I think something like this would work though if you want go this route instead: https://github.com/benfred/rbspy/commit/749938a81239f03301e8abb41ddd04200dcbe587

src/core/proc_maps/linux_maps.rs
benfred: Good point - making the implied interface explicit seems like a great idea!

I’ve added here: https://github.com/rbspy/rbspy/pull/166/commits/bfe6ad2e8e0f54fe64d9f39f9d6faa4d3cde73a9 but got a little hung up on naming the trait. I went with IMapRange as a placeholder (so as not to conflict with the MapRange structs) but I am totally open to better suggestions.

benfred:
I've split of the [proc_maps code to an external crate](https://github.com/benfred/proc-maps) since creating this PR (and got windows working with rbspy =). 

I'm also not sure if I broke things with that last commit to create an explicit interface (the unittests all pass on my local machine  - but seem to fail here =(. I've left the commit for the explicit interface out of the proc-maps crate for now anyways.

I'm leaving this PR open for the discussion, but I think most of the changes should happen in the proc-maps crate or the windows PR #165 now - so I don't think we should merge this

jvns:
closing sounds good to me!

PR #170, opened 2018-08-27
Omit MacOS root warning for rbspy report

wabain:
This PR contains a few distinct changes, bundled together because the code they touch overlaps. (I can break things out/back parts out if needed.)

* Remove some unused dependencies
* Don't require the nix package on Windows (all uses are already in `cfg(unix)` blocks)
* More substantively: previously, a warning was printed on MacOS if `rbspy report` was invoked by a non-root user, even though that command doesn't actually require root and works without it. Reorder things so that the test for root is only executed when needed.
* Finally, print the non-root warning to stderr instead of stdout

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7,14 +7,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "aho-corasick"
-version = "0.6.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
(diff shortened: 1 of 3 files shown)


jvns:
Thanks! Nice catch with the `mach` etc dependencies -- those were pulled out into the `proc_maps` crate recently and looks like we forgot to remove the dependency.

Issue #145, opened 2018-04-30
os error 14 on Ubuntu


pixelastic:
Hello,

I tried to use rbspy to profile a `jekyll build` call on Ubuntu 16.04, but I'm having the following error:

```
$ rbspy record -- bundle exec jekyll build
Press Ctrl+C to stop
Dropped 29/57 stack traces because of errors.
Wrote raw data to /home/tim/.cache/rbspy/records/rbspy-2018-04-30-J3YBZ0ZThp.raw.gz
Writing formatted output to /home/tim/.cache/rbspy/records/rbspy-2018-04-30-73lvDNz4eP.flamegraph.svg
Error. Causes:
- Failed to copy memory address 7fd85f24c820
- Bad address (os error 14)
```

The actual `jekyll build` command is correctly executed, though.

The repository to reproduce the error can be found here: https://github.com/algolia/jekyll-algolia-example
```
$ bundle install
$ rbspy record -- bundle exec jekyll build
```

Happy to help if there's anything I can do :)

jvns:
Thanks for reporting this! I tried it on my computer and got [this flamegraph](https://user-images.githubusercontent.com/817739/39461458-bf16ed00-4cd8-11e8-9d86-520625d259d4.png), so I don't think it's to do with the Ruby code here.

How did you install Ruby? It would help if you could send me a core dump or two of the Ruby process's memory so I can see if I can figure out what's going on. Here's how to get a core dump:

first, start a while loop that will try to grab a core dump:
```
while true
do 
sudo gcore $(pgrep -f jekyll)
sleep 0.5
done
```

and then run `bundle exec jekyll build`

`gcore` will generate a file named like `core.24996`, which is the core dump. 

you can either upload them somewhere or email them to julia@jvns.ca.

pixelastic:
I'm using ruby through `rvm`. Here is the output of `ruby -v`:
```
ruby 2.4.2p198 (2017-09-14 revision 59899) [x86_64-linux]
```

Here is the incomplete flamegraph I initially got: 
![rbspy-2018-05-02-yzldx6arj9 flamegraph](https://user-images.githubusercontent.com/283419/39516568-9a6fe620-4dfd-11e8-9605-ef4a9cd33e59.png)

And here are two dumps:
- http://tmp.pixelastic.com/share/core.19844
- http://tmp.pixelastic.com/share/core.21121

Thanks!

pathawks:
[...]

PR #177, opened 2018-09-04
Handle execve's on Linux

wabain:
This is a fix for #145, handling `exec` calls, for Linux. While the same approach should work for MacOS, extra work would be needed to distinguish between invalid reads and process exit and since I've already let this sit for many months, I'd prefer to land Linux support.

The idea here is to check the result returned when getting the stack trace from the target process to detect the error that the pointer to the ruby thread state, specifically, has become invalid. If it has, we then reinitialize StackTraceGetter on the basis that if we've lost the thread the whole process is invalid.

In the test case I added I'm polling until the reinitialization happens. I did this because when I used a fixed timeout I was sometimes getting the trace before reinitialization (one every 500 test runs or so), and I didn't want to play the timeout tweaking game. With the current polling I still get the error "stack base and cfp address out of sync" once every few thousand test runs but I think that error should be transient and shouldn't impact real usage.


neosilky:
Looks good to me and appears to fix #145 :+1:

Issue #188, opened 2018-09-26
Include current sample rate when rbspy fails to sample at that rate


janpio:
What does that mean exactly?
Is this a nice tip to get a little bit better results, or are my current results total crap and only using this will give me something useful?

janpio:
I should have mentioned that: Whatever the default is. I didn't configure a thing, then got this scary output.

jvns:
re "what does this message mean"  -- if the percentage is low (like 3.6%), you don't have to worry and the results are fine.

jvns:
The error message:

> "3.6% (151/4217) of stack traces were sampled late because we couldn't sample at expected rate, results may be inaccurate. Try sampling at a lower rate with `--rate`."

should include the current sample rate.

jvns:
Tagging this as "good first issue" because including the sample rate in this error message should be really easy :)

PR #191, opened 2018-10-09
Add current sample rate to print summary

at1as:
This PR adds the `sample rate` to the print summary when the current rate is lower than the setting. 

This is in response to: https://github.com/rbspy/rbspy/issues/188

Let me know if the verbiage should be changed, or if there's anything else I've missed. I'm a complete beginner with Rust 😅


jvns:
Thanks so much for this! Apologies for taking a while to get to it :)

PR #194, opened 2018-11-02
Add ruby 2.3.8 support

printercu:
Hi! I'm not sure if this enough to add support for 2.3.8. Please advise, if this requires some more changes.


jvns:
this doesn't compile because it's missing the generated ruby 2_3_8 module (generated by scripts/bindgen.sh). I opened https://github.com/rbspy/rbspy/pull/195 which has the required generated code. thanks for bringing this up!

PR #199, opened 2019-02-04
Sample thread_ids

ViugiNick:
We want to use `rbspy` as a profiler engine for RubyMine. It turned out it turned out that dividing samples not only by pids but by thread_id quite usefull. As in the case of rails for example some threads are always loop some library code, such threads add some noise to a flame graph with mixed thread. In addition, it allows user to better localize the problem, as only the flamegraph of the desired thread is considered (with the needed method, for example)

screenshot from 2019-02-04 22-15-37

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -382,7 +382,7 @@ dependencies = [
 
 [[package]]
 name = "rbspy"
-version = "0.3.2"
+version = "0.3.3"
 dependencies = [
  "byteorder 0.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


ViugiNick:
@ViugiNick Thanks, in the near future I will add a PR with "silent" mode. Or is it already possible to disable every second logging?

jvns:
what do you mean by silent mode?

ViugiNick:
@jvns I mean the ability to turn off printing of the summary profiling data in stdout every second

jvns:
that's a good idea, it's not possible today! it might also be interesting
to look at py-spy's user interface (https://github.com/benfred/py-spy) for
ideas!

PR #200, opened 2019-02-11
Add --silent flag, which allows to cancel every second logging

ViugiNick:
As we discussed in https://github.com/rbspy/rbspy/pull/199, sometimes it is needed to turn off the logging of the profiler's summary data.

P.S Can you please release the latest rpspy version, so we can share a link to it?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -382,7 +382,7 @@ dependencies = [
 
 [[package]]
 name = "rbspy"
-version = "0.3.3"
+version = "0.3.5"
 dependencies = [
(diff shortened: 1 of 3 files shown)


ViugiNick:
@jvns Could you please take a look?

jvns:
Thanks for the ping - this looks great.

ViugiNick:
@jvns Thanks, it would be very nice to release a new version of the profiler.

jvns:
i pushed a 0.3.5 tag which should do it.

ViugiNick:
@jvns Could you please update homebrew too?

PR #211, opened 2019-05-17
WIP: Replace original FlameGraph with Rust port

neosilky:
[Inferno](https://github.com/jonhoo/inferno) is a Rust port of the original FlameGraph, mainly focussing on performance.

The conversion to using Inferno was super easy and provides the same functionality. This would get rid of the requirement for perl to be installed and means we can get rid of the vendored code.

However, there currently seems to be a bug where an extra stack line is added. I'm not sure why but I'm looking into it when I get a chance.


benfred:
@neosilky - nice work on this! your change inspired me to try to do the same with py-spy.

I figured out that the extra stack line is because of a trailing ';' when building up the frame ([this line on the final iteration](https://github.com/rbspy/rbspy/blob/master/src/ui/flamegraph.rs#L30)) - which seems to make a difference with inferno but not so much with the vendored flamegraph code.

I have a PR with py-spy https://github.com/benfred/py-spy/pull/114 that fixes this - and also doesn't write out a temporary file (using inferno from_lines to create). Feel free to copy anything you find useful from it, the flamegraph code in py-spy was copied from rbspy originally so should be basically the same. Eventually I'd like to merge these into a common crate (when I some more free time)

jonhoo:
It should also be pretty easy to make inferno ignore a trailing `;` -- I'd be happy to review a PR :)

neosilky:
@benfred Nice, thanks! Will update my code when I get a chance :)

neosilky:
Now I'm getting bad coloring :grin: 

image

jvns:
This is great, will merge once the coloring issue is sorted out :)

jonhoo:
That coloring looks like you are rendering a [differential flame graph](http://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html). Inferno should only be generating that if you give it folded (i.e., post-stackcollapse) files that have _two_ sample counts at the end of each line. Is that the case?

neosilky:
@jonhoo Ah, I switched to joining each frame by a `;` but now the end of the output is like `...;<c function> - unknown line 0 1`, which looks like the differential flamegraph format you mention.

So maybe the easiest thing here is to get Inferno to ignore final `;`s.

neosilky:
@jvns I can easily fix it by switching the line format to `aaa - this/is/my/file.rb:32`. Is that ok? Or do you want to keep "`line`" in there?

neosilky:
(Woo, pure Rust!)

image

jvns:
this looks great! are there more changes you want to make before merging?

neosilky:
@jvns Should be good to go once these tests pass :)

jonhoo:
@neosilky ah, I see, you specifically have a "function" in the stack trace that contains spaces, is last, and ends with a number -- yeah, that would cause some oddity. I guess the trailing `;` is what saved you from that parsing ambiguity in the old version. Adding this to inferno seems like a good idea regardless, so happy to add this from a PR :)

PR #219, opened 2019-06-20
Downgrade nix

neosilky:
Downgrade Nix to fix build


jvns:
what's the reason for this?

>

neosilky:
@jvns I merged a PR earlier which passed at the time but now it's failing on CI. Works fine locally but it looks like downgrading Nix fixes CI. Not sure why but investigating.

PR #246, opened 2020-01-27
Allow merging multiple reports together

liaden:
This changes `rbspy report` subcommand:
* Allow multiple inputs to be specified with `-i`.
* Add `-m` flag to allow merging multiple inputs into one output. 
* Let `-o` for output be optional. 
  - It automatically applies a default name: 
    * When merging, this defaults to `"merged" + extension`.
    * Without merging, it takes the root from the input it is processing. 
  - Output filename does not require extension being supplied.

This changes `rbspy record` subcommand:
* Both `--raw-file /tmp/raw` and `--raw-file /tmp/raw.raw.gz` will set the filename to `raw.raw.gz`.
* It will fix extensions for formats: `--file flamegraph.txt --format flamegraph` will set filename to `flamegraph.svg`. No supplied will append it correctly.
* `--format raw` will only output the raw file. This avoids the post processing step that can be more cpu expensive when left running long enough at a high enough rate.

 Testing:
- [x] Some manual testing.
- [x] Merged report.
- [x] Multiple reports and multiple outputs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -842,7 +842,7 @@ dependencies = [
 
 [[package]]
 name = "rbspy"
-version = "0.3.9"
+version = "0.3.10"
 dependencies = [
  "byteorder 1.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "chrono 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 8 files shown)


acj:
Closing due to age and conflicts, but feel free to reopen this if you're still interested.

PR #248, opened 2020-02-03
Add proof of concept start/stop profiling mechanism

### LINKED ISSUE #247: Enable and disable profiling from Ruby program
To help isolate program segments for profiling, it would be helpful to be able to start and stop profiling from within the Ruby program (such as pausing and resuming profiling in ruby-prof).

brunsa2:
Closes #247.

Proof of concept starting/stopping profiling via Ruby program sending SIGUSR1 and SIGUSR2. Tested on macOS so far.

This is my first Rust project so if this would be helpful and there is a better way to do this (better in the Rust code, different IPC mechanism, etc.), please let me know. I'd still like to add support for other platforms, add arguments to enable/disable this feature, add a Ruby gem for ease in sending the signal, and ensure that I'm using the `AtomicBool`s correctly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,6 +22,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "arc-swap"
+version = "0.4.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "arrayvec"
 version = "0.4.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


jvns:
I love that this is so simple! My main concern is that (as a user) I'd expect for a "stop profiling" signal to actually stop rbspy's collection of stack traces (which can use a significant amount of CPU, especially at a high sampling rate).

So I'd prefer for this to actually make the stack-trace-collecting thread stop collecting stack traces instead of just making the consumer of those stack traces stop printing them.

Here's call to the function that gets stack traces (which is what I think we should pause): https://github.com/rbspy/rbspy/blob/ab791cda4f71a394ebd8e94a98f5f3151012d2a8/src/main.rs#L596

PR #249, opened 2020-02-11
Add ruby 2.7.0 support

liaden:


src/core/ruby_version.rs
```diff
@@ -138,6 +138,26 @@ macro_rules! ruby_version_v2_6_x(
         }
         ));
 
+macro_rules! ruby_version_v2_7_x(
```
liaden: I did not change this relative to `ruby_version_v2_6_x`. Verified that I still got a flamegraph on ruby 2.7 after building.

jvns: If it's the same as `ruby_version_v2_6_x` maybe it would be clearer to rename it to `ruby_version_v2_6_to_2_7` or something? (like `ruby_version_v_2_3_to_2_4`). Thanks for doing this.

igorwwwwwwwwwwwwwwwwwwww:
@liaden Thanks for this PR!

It looks like support for 2.7.0 has been merged in the mean time: https://github.com/rbspy/rbspy/pull/256. So I think we can close this one. If you have any questions, let me know!

PR #253, opened 2020-04-06
add timestamp weights to speedscope output format

### LINKED ISSUE #252: Update speedscope output to include unit & timestamp
Thanks to @igorwwwwwwwwwwwwwwwwwwww's work in #250, timestamps are now included in profiles! This means that profiles exported to speedscope can be more accurate and include units to make the profiles easier to understand.

I think the relevant changes needed are to change these two lines:
 https://github.com/rbspy/rbspy/blob/5bfe998f5ecfe5929ea9b5494aa340b5ea49b09e/src/ui/speedscope.rs#L114

https://github.com/rbspy/rbspy/blob/5bfe998f5ecfe5929ea9b5494aa340b5ea49b09e/src/ui/speedscope.rs#L121 

I'm not sure when I'll get around to doing this since I don't work in ruby very often, but if someone is excited about this, I'm happy to answer any questions about speedscope, and I don't think it would be a very challenging change to make.

igorwwwwwwwwwwwwwwwwwwww:
This patch adds weights to the speedscope profile as proposed by @jlfwong.

It drops the first sample by assigning a weight of 0 (since we have no delta to compare to) and then assigns weights by comparing the timestamp to the previous one.

I don't have much experience with rust, so if there's any hints on how to make some of this nicer, I'd love to get some pointers!

Closes #252.

### DIFF
--- a/src/ui/speedscope.rs
+++ b/src/ui/speedscope.rs
@@ -2,6 +2,7 @@ use std::collections::{HashMap};
 use std::io;
 use std::io::Write;
 use std::fs::File;
+use std::time::SystemTime;
 
 use crate::core::types::{Pid, StackTrace, StackFrame};
 
@@ -97,7 +98,7 @@ enum ValueUnit {
 }
 
 impl SpeedscopeFile {
-  pub fn new(samples: HashMap<Option<Pid>, Vec<Vec<usize>>>, frames: Vec<Frame>) -> SpeedscopeFile {
+  pub fn new(samples: HashMap<Option<Pid>, Vec<Vec<usize>>>, frames: Vec<Frame>, weights: Vec<f64>) -> SpeedscopeFile {
     let end_value = samples.len();
 
     SpeedscopeFile {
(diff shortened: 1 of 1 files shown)


igorwwwwwwwwwwwwwwwwwwww:
Note: I tested this by uploading a 10 second profile to speedscope, and speedscope indeed displayed a 10 second timeline.

@jlfwong it'd be great if you could double check this and maybe also give it a test run, since I'm not super familiar with the speedscope side of things.

igorwwwwwwwwwwwwwwwwwwww:
@jlfwong another ping :)

jlfwong:
Oh, hi, sorry I missed this! I don't actually do ruby dev *or* rust dev, so I'd probably just run `rbspy` on [this file from the speedscope repo](https://github.com/jlfwong/speedscope/blob/master/sample/programs/ruby/simple.rb), which is I guess about the same kind of verification you've already done.

I also don't have commit privileges on this repo, so you'll need someone else to look at it regardless.

igorwwwwwwwwwwwwwwwwwwww:
Okay, in this case I think we're good to go on this PR. My tests indicated that this is working. @jvns ping! :)

jvns:
done, thanks so much for pinging about this!

PR #264, opened 2020-06-14
Add support for flamegraph intermediate output

paul-mannino:
Add support for outputting trace data into the folded
format that can be turned into a flamegraph.svg by
flamegraph.pl/inferno-flamegraph/etc.

Testing... Lots of different opinions about when/how IO should be tested--wanted to check in before I went nuts writing integration tests.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -83,6 +83,7 @@ arg_enum!{
     #[allow(non_camel_case_types)]
     pub enum OutputFormat {
         flamegraph,
+        flamegraph_stack_lines,
         callgrind,
         speedscope,
         summary,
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -83,6 +83,7 @@ arg_enum!{
     #[allow(non_camel_case_types)]
     pub enum OutputFormat {
         flamegraph,
+        flamegraph_stack_lines,
```
paul-mannino, line 86: Struggled with naming here because the original project was kind of fuzzy about what it called this state. Maybe flamegraph_raw or raw_flamegraph is better?

acj:
Hey @paul-mannino, thanks for the PR, and sorry for the long delay in getting it reviewed. I'm happy to work with you to get this freshened up and merged if you're still interested in the feature.

re: testing, I think something in the style of the existing flamegraph test should work. Feel free to suggest other testing ideas, though.

src/main.rs
acj: I think a name with "raw" makes sense. flamegraph_folded also seems reasonable.

michelhe: Hey, I was wanting to work on adding this to rbspy and found out that this PR exists. Thanks :) 
I believe that a more suitable naming for this would be `collapsed` or `raw-collapsed`

michelhe:
@acj Any plans to merge this PR in the near releases?

acj:
@michelhe If there's interest, I'm fine with cleaning this up and merging it. Would you be interested in opening a fresh PR?

michelhe:
@acj Sure, no problem.

acj:
Closing in favor of #326. Thanks for getting this started, @paul-mannino!

PR #270, opened 2020-11-05
add Dockerfile for use with bindgen.sh

igorwwwwwwwwwwwwwwwwwwww:
This allows you to run a bindgen in a relatively self-contained environment. Some hackery was needed to get it working, and it doesn't generate exactly the same files that exist in the repo, at least not yet. This is quite possibly due to differing bindgen versions, and possibly other updates in rust packages.

I had to go with an older debian version because older rubies require an older version of bison (https://bugs.ruby-lang.org/issues/17106).

I've opened a separate MR for `2_7_2`, which we don't have support for yet: https://github.com/rbspy/rbspy/pull/271.

Issue #267, opened 2020-09-17
Support for Ruby 2.5.2?


Pr0p1k:
Is there some critical bug in this version, so it's not supported by RbSpy?

igorwwwwwwwwwwwwwwwwwwww:
closed by https://github.com/rbspy/rbspy/pull/272

PR #272, opened 2020-11-06
add support for ruby 2.4.10 and 2.5.2

### LINKED ISSUE #258: Support for Ruby 2.4.10
Please add support for Ruby 2.4.10

Thanks

igorwwwwwwwwwwwwwwwwwwww:
Both of these were tested against `./ci/ruby-programs/infinite.rb` on macOS and are working.

Both were generated via the bindgen Dockerfile from https://github.com/rbspy/rbspy/pull/270.

closes #258, #267


jvns:
lgtm!

the steps to make a release are:

1. make a change to Cargo.toml with the new version ID
2. tag that commit with the new version (`git tag whatever`)
3. push the tags & commit to github (`git commit --tags`)

(this should probably be documented I guess?)

The Travis build is responsible for automatically publishing the new release tarballs to github when the new tag is pushed. I'm a bit worried that eventually we'll run into problems with Travis because of their reduced support for open source projects, but Travis seems to be more or less working right now.

jvns (approved):

Issue #275, opened 2020-11-26
Allow to set the `min_width` option from CLI to increase/reduce FlameGraph precision/fidelity


jarthod:
I noticed that no matter how long I let `rbspy` run (e.g. 2-4 hours), many flames with more than 1000 samples were still missing from the SVG flamegraph. I found the `min_width` [option from inferno](https://docs.rs/inferno/0.10.1/inferno/flamegraph/struct.Options.html#structfield.min_width) as mentioned in #240 (by @liaden) which is currently set to the default of `0.1%`. I think this is a good default value to keep a decently sized output.

It could be useful for people like me who run long profiling (and so have many sample so good precision) to be able to decrease this number to have more details in smaller flames (I would probably want to go down to 0.01% if it doesn't crash my browser when opening the SVG ☺)

Has adding this option been discussed already? (sorry if I couldn't find the issue) would it be ok to add it? If so I can try sending a PR but I don't promise anything as I only tried rust a long time ago ;)

jvns:
I don't think it's been discussed before, but this sounds reasonable -- I'd be happy to merge a PR that does this.

jarthod:
Opened PR #276 :tada:

PR #276, opened 2020-12-02
Add flamegraph precision option to control min_width

jarthod:
Following suggestion #275, here is my first rust PR :tada: 

Ok so the idea is to be able to control the `min_width` [option from inferno](https://docs.rs/inferno/0.10.1/inferno/flamegraph/struct.Options.html#structfield.min_width) which is currently kept to the default of `0.1%`. It could be useful for people like me who run long profiling (and so have many sample so good precision) to be able to decrease this number to have more details in smaller flames.

So in this PR I tried to add a `flamegraph-precision` option, which is an integer (I couldn't manage to put a float in the `Record` struct), defined as:

```
> rbspy record --help
...
OPTIONS:
...
        --flamegraph-precision <PRECISION>    Minimum flame width = 1/PRECISION % [default: 10]
```

So default value of 10 = 0.1%, setting 1 would give 1% (lower precision) and setting 100 would give 0.01% (higher precision) for example.

I tested this locally and was able to successfully generate more or less detailed flamegraphs. I added an automated test for the parameter parsing which passes:

```
> cargo test test_arg_parsing

... skipping all warning ...

    Finished test [unoptimized + debuginfo] target(s) in 0.05s
     Running target/debug/deps/rbspy-57c9c1ccd82df402

running 1 test
test tests::test_arg_parsing ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 18 filtered out
```

The other tests are not all running on my machine (Ubuntu 20.04) and one hangs the suite (with master) but that is an unrelated matter. The test suite is green on Travis :+1: 

I am not sure if I passed the argument the right way, let me know if you think I should rewrite this another way.
I let the default value of 0.1% for the `report` command, let me know if I should also add the parameter to this command too!

Thanks :bow: and sorry in advance if I wrote any rust monstrosity :see_no_evil:

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -88,7 +88,7 @@ arg_enum!{
 }
 
 /// Subcommand.
-#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+#[derive(Clone, PartialEq, PartialOrd, Debug)]
 enum SubCmd {
     /// Record `target`, writing output `output`.
     Record {
@@ -101,7 +101,7 @@ enum SubCmd {
         no_drop_root: bool,
         with_subprocesses: bool,
         silent: bool,
-        flamegraph_precision: u32
+        flamegraph_precision: f64
     },
     /// Capture and print a stacktrace snapshot of process `pid`.
     Snapshot { pid: Pid },
@@ -110,7 +110,7 @@ enum SubCmd {
 use SubCmd::*;
 
 /// Top level args type.
-#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+#[derive(Clone, PartialEq, PartialOrd, Debug)]
 struct Args {
     cmd: SubCmd,
 }
@@ -301,7 +301,7 @@ fn snapshot(pid: Pid) -> Result<(), Error> {
 }
 
 impl OutputFormat {
-    fn outputter(self, flamegraph_precision: u32) -> Box<dyn ui::output::Outputter> {
+    fn outputter(self, flamegraph_precision: f64) -> Box<dyn ui::output::Outputter> {
         match self {
             OutputFormat::flamegraph => Box::new(output::Flamegraph(ui::flamegraph::Stats::new(flamegraph_precision))),
             OutputFormat::callgrind => Box::new(output::Callgrind(ui::callgrind::Stats::new())),
@@ -538,7 +538,7 @@ fn parallel_record(
     silent: bool,
     sample_rate: u32,
     maybe_duration: Option<std::time::Duration>,
-    flamegraph_precision: u32,
+    flamegraph_precision: f64,
 ) -> Result<(), Error> {
 
     let maybe_stop_time = match maybe_duration {
@@ -673,7 +673,7 @@ fn record(
 fn report(format: OutputFormat, input: PathBuf, output: PathBuf) -> Result<(), Error>{
     let input_file = File::open(input)?;
     let stuff = storage::from_reader(input_file)?.traces;
-    let mut outputter = format.outputter(10);
+    let mut outputter = format.outputter(0.1);
     for trace in stuff {
         outputter.record(&trace)?;
     }
@@ -832,8 +832,8 @@ fn arg_parser() -> App<'static, 'static> {
                         .required(false)
                 )
                 .arg(
-                    Arg::from_usage("--flamegraph-precision=[PRECISION] 'Minimum flame width = 1/PRECISION %'")
-                        .default_value("10"),
+                    Arg::from_usage("--flamegraph-precision='Minimum flame width in %'")
+                        .default_value("0.1"),
                 )
                 .arg(Arg::from_usage("<cmd>... 'command to run'").required(false)),
         )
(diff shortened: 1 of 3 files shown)


acj:
Hey @jarthod, thanks for this, and sorry for the delay. The changes look pretty reasonable to me. I'll try it out this weekend and will let you know if I have any feedback. Would you mind rebasing the branch so that CI can run again?

jarthod:
@acj sure no problem! I just rebased the branch and resolved some conflicts regarding default options which changed (let me know if this is not correct). The specs are green!

src/main.rs
```diff
@@ -827,6 +831,10 @@ fn arg_parser() -> App<'static, 'static> {
                     Arg::from_usage( "--silent='Don't print the summary profiling data every second'")
                         .required(false)
                 )
+                .arg(
+                    Arg::from_usage("--flamegraph-precision=[PRECISION] 'Minimum flame width = 1/PRECISION %'")
```
acj, line 835: I find this somewhat confusing to read. The "precision" abstraction is nice -- higher number gives you greater resolution in the call stacks, and lower gives you less precision. Mentally converting the precision into a width isn't intuitive, though (at least to me).

There may be a clean solution. You mentioned in the PR description that you tried to use a float type for the parameter but couldn't add it to the `Record` struct. I think we can unblock that by removing the `Eq`, `Ord`, and `Hash` traits from the struct. Some edge case bugs around float comparisons will be possible with that struct, but we only compare them in the tests (mostly to check the strings), so I think we'll be okay. If you remove those traits and switch the parameter types to `f64`, everything should work. And then rbspy will be consistent with other tools that use a "minimum width" argument like the original flamegraph library does.

Here's a diff of the key change:

```diff
--- a/src/main.rs
+++ b/src/main.rs
@@ -88,7 +88,7 @@ arg_enum!{
 }
 
 /// Subcommand.
-#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+#[derive(Clone, PartialEq, PartialOrd, Debug)]
 enum SubCmd {
     /// Record `target`, writing output `output`.
     Record {
@@ -110,7 +110,7 @@ enum SubCmd {
 use SubCmd::*;
 
 /// Top level args type.
-#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+#[derive(Clone, PartialEq, PartialOrd, Debug)]
 struct Args {
     cmd: SubCmd,
 }
@@ -301,7 +301,7 @@ fn snapshot(pid: Pid) -> Result<(), Error> {
 }
 
```

acj (changes requested):
Okay, I had a chance to try this and to develop some feedback. The parameter will be very nice to have. A couple of things to resolve, and then I'll be happy to merge. Thanks again for your patience.

src/main.rs
```diff
@@ -827,6 +831,10 @@ fn arg_parser() -> App<'static, 'static> {
...
+                        .default_value("10"),
```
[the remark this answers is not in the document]
jarthod: Good catch, indeed I think `0.1` is still a good default to produce decently sized flamegraphs, `0.01` could easily lead to very big files which are hard to load on some machines. So I tend to agree with keeping 0.1 as default value.

src/main.rs
jarthod: Ah yes, indeed I wasn't sure what these traits were used for so I didn't even try to remove them, but you're right using a float is better here, I just applied these changes in d6bdf84fcf6548de6c5ecc3c7efff756afc780aa but what do you think about the parameter name then? I suppose "flamegraph-precision" doesn't make much sense now so do you have any recommendation? I'm thinking maybe "flame-min-width"?

acj: flame-min-width sounds good to me. It's reasonably short (easy on the old fingers), and its purpose should be clear enough to anyone who decides to use it.

jarthod:
> Okay, I had a chance to try this and to develop some feedback. The parameter will be very nice to have. A couple of things to resolve, and then I'll be happy to merge. Thanks again for your patience.

Thanks! I commented inline with a first version. Also I have a question about testing, I wrote a small ruby script (like the ones in `ci/ruby-programs`) to be able to easily verify this new parameter is working:
```ruby
def recurse_sleep(seconds)
  return if seconds <= 0.00001
  sleep(seconds/2.0)
  recurse_sleep(seconds/2.0)
end

recurse_sleep(30)
```

It gives flamegraphs like this:

min-width = 1%:
image

min-width = 0.1% (default):
image

min-width = 0.02%:
image

But now I'm wondering if it's worth adding it to the repo for automated testing? the unit test I copied about argument parsing was easy enough but testing this for real and verifying the width of the smallest flame looks a bit more complicated to do in rust ^^ So let me know if you think:
- We should just forget about this
- I should commit the ruby script (for manual testing at least)
- Or I should try to write a test for this case

src/main.rs
jarthod: All right! I just pushed the rename in dbcb6a80813dd9285f463be854ff513e42ea5261. Thanks!

acj:
I think the script will be useful, so my vote would be to commit it but not use it in the automated tests (at least for now). As for the tests themselves, we would mostly be testing the inferno library and not rbspy code if we tried to verify the flame widths. The argument parser test you've already added seems fine to me.

The most recent changes LGTM. Feel free to add the script if you want. Once you're happy with the PR, I'll go ahead and merge. This will go out with the 0.5 release, probably tonight or tomorrow.

acj (approved):

jarthod:
Perfect, thanks! I just pushed the script in 21136f2ab99458941fb9c0e7c9f6bfb9805c7083

acj:
Thanks!

PR #280, opened 2021-02-06
macOS M1 support - get_dyld_info works better than get_process_maps

### LINKED ISSUE #279: macOS M1 support
Hi everyone!

I tried running rbspy on an M1 mac and it did not work, so I did some debugging. I have not fixed the issue yet, but I thought this could be helpful to anyone who might be looking at this in the future.

When you try to use rbspy on an M1 mac it fails with this error: `Ruby version not supported yet: . Please create a GitHub issue and we'll fix it!`. So it's pretty apparent that the version is not being read correctly.

So I looked at how the version was being detected and I was able to trace it down all the way from `rbspy` to `proc-maps` to [libproc-rs](https://github.com/andrewdavidmackenzie/libproc-rs).

### Summary

It looks like `libproc::libproc::proc_pid::regionfilename` and `proc_regionfilename` are returning incorrect names, here's the relevant code: https://github.com/andrewdavidmackenzie/libproc-rs/blob/d8d1c40dbd9d8cf9bc70314690c4a38180cdcc23/src/libproc/proc_pid.rs#L336

And as a result the memory region for libruby is not being detected correctly.

I added some code [here](https://github.com/rbspy/proc-maps/blob/master/src/mac_maps/mod.rs#L137) in `proc-maps` to print filenames and addresses and then compared it to `vmmap` output.

Pasted_Image_2_4_21__10_13_PM

[here's a gist with this data in text form](https://gist.github.com/petethepig/e6b7c39f6a84def0d963615ae0fb0fb9)

### Next Steps

* [ ] find a way around this? maybe there's some other method we could use to filter these out / detect `regionfilename` correctly. 
* [ ] file a bug with Apple? It looks like `proc_regionfilename` is just returning incorrect information. I would love any help / pointers on this as I have never dealt with this before.

### Other Relevant Information

* This is only relevant for ruby running with **Rosetta**, I did not try the native version yet.
* I'm not sure if this is the only hurdle on the way to M1 support.

petethepig:
Adds macOS M1 support, fixes #279, there's more context in the issue.

### DIFF
--- a/src/core/address_finder.rs
+++ b/src/core/address_finder.rs
@@ -26,8 +26,7 @@ mod os_impl {
     use crate::core::initialize::IsMaybeThreadFn;
     use crate::core::types::Pid;
 
-    use proc_maps::{get_process_maps, MapRange};
-    use proc_maps::mac_maps::{get_symbols, Symbol};
+    use proc_maps::mac_maps::{get_symbols, Symbol, get_dyld_info, DyldInfo};
 
     use failure::Error;
 
@@ -102,7 +101,7 @@ mod os_impl {
     }
 
     fn get_program_info(pid: Pid) -> Result<ProgramInfo, Error> {
-        let maps = get_process_maps(pid).map_err(|_| {
+        let maps = get_dyld_info(pid).map_err(|_| {
             AddressFinderError::MacPermissionDenied(pid)
         })?;
         let ruby_binary = get_ruby_binary(&maps)?;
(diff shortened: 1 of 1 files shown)


src/core/address_finder.rs
```diff
@@ -113,28 +112,18 @@ mod os_impl {
         })
     }
 
-    fn get_ruby_binary(maps: &Vec<MapRange>) -> Result<Binary, Error> {
-        let map: &MapRange = maps.iter()
-            .find(|ref m| if let Some(ref pathname) = m.filename() {
-                pathname.contains("bin/ruby") && m.is_exec()
```
petethepig, line 119: I have some concerns about `m.is_exec()` — there's no such field on `DyldInfo`... not sure how important this part is.

jvns: I don't remember if I put that there for a reason or not, so this seems fine

jvns:
I don't have any way to test this but it sounds like you've tested it so I'll merge it and make a release.

PR #283, opened 2021-02-09
Try out using github actions for CI

jvns:
Going to push a billion changes to this branch

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,114 @@
+# originally copied from https://github.com/BurntSushi/ripgrep/blob/master/.github/workflows/ci.yml
+name: ci
+on:
+  pull_request:
+  push:
+    branches:
+    - master
+  schedule:
+  - cron: '00 01 * * *'
+jobs:
+  test:
+    name: test
+    env:
+      # For some builds, we use cross to test on 32-bit and big-endian
+      # systems.
+      CARGO: cargo
+      # When CARGO is set to CROSS, this is set to `--target matrix.target`.
+      TARGET_FLAGS:
+      # When CARGO is set to CROSS, TARGET_DIR includes matrix.target.
+      TARGET_DIR: ./target
+      # Emit backtraces on panics.
+      RUST_BACKTRACE: 1
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        build:
+        - stable
+        - macos
+        - win-msvc
+        - win-gnu
+        include:
+        - build: stable
+          os: ubuntu-18.04
+          rust: stable
+        - build: macos
+          os: macos-latest
+          rust: stable
+        - build: win-msvc
+          os: windows-2019
+          rust: stable
+        - build: win-gnu
+          os: windows-2019
+          rust: stable
+    steps:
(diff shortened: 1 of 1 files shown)


benfred:
I also took a bit of a look at getting github actions going for rbspy last night =) You're farther along here than I got to - but I thought I'd share my attempts anyways.

 I saw the PR that fixed the aarch64 build - and wanted to see if we could build armv7/aarch64 with github actions.    I had some problems running the tests, but cross compiling for 4 different processors with musl worked: https://github.com/benfred/rbspy/blob/7329d2e4e1028be9d7aa32bf1c94aaa4103430c0/.github/workflows/build.yml#L48-L66
This uses the same steps as in py-spy - using the [rust-musl-cross](https://github.com/messense/rust-musl-cross) project to provide docker images for each target.  

Including the cross compile targets for armv7/aarch64/i686 etc might be worth including here - it'd be good to check all the arm builds.

jvns:
Oh, that's awesome! @benfred want to make a PR with just the linux parts of that build and then we can try to sort out the macos/windows build after? Having arm linux support would already be a big improvement. 

I'm not sure what to do about the failing test on Mac yet -- my hypothesis right now is that somehow the github actions runner doesn't have the permissions it needs to actually read the memory from another process (the failing test is an integration test where it actually runs a ruby program and tries to rbspy it). 

Does py-spy have an integration test like that? Does it work in GitHub actions?

benfred:
thanks! I'll submit a PR with the linux cross stuff later tonight 

For mac - I *think* you will need to run as root.  When I was setting up github actions, I had to split the mac tests off from the linux/windows ones - and use sudo for running the mac tests here: https://github.com/benfred/read-process-memory/blob/4e783fe61b431180b40559bcb4524d2cc42ad846/.github/workflows/build.yml#L22-L27

(py-spy needs to get fixed up to do something similar for mac, but isn't doing it yet - the mac tests aren't running fully right now).

benfred:
I made a PR with those changes here (linux cross compile / test mac as root) https://github.com/rbspy/rbspy/pull/284 - it looks like its all good.

For the linux cross compile, I set up to upload the output as github actions artifacts.  I did something similar with py-spy - and then when we tagged a release - those artifacts automatically get both uploading to pypi and to the github releases page:  https://github.com/benfred/py-spy/blob/b9fab61ac5869888157094f130456ae9ceb1a5b2/.github/workflows/build.yml#L82-L105 . Not strictly necessary, but might be nice for people to be able to get the linux binaries for their platform.

We could also do the same for the mac build (and maybe windows in the future?) so that the releases page has them all set up.

jvns:
@benfred Can you take a quick look at my last commit before I merge this? I just modified the macos/linux build so that it'll also upload a release artifact, and I switched the tests to use `--release` so that can reuse the build compilation and go faster.

I really like that this uploads an artifact for every PR -- it feels like it could make it way easier for folks to try out a proposed change from a PR. I just wish that there was a direct link to to the artifact from the PR page -- I wonder if there's a way to do that.

benfred:
this looks great!  thanks!

I don't know of a way to link the artifacts direct from the PR (though I'm still just learning github actions, so there might be something out there).

PR #282, opened 2021-02-09
Resolve C function names when sampling

acj:
As noted in #110, rbspy currently lumps all C functions into a single bucket. It would be very useful to separate them so that we can understand which C functions are consuming the most time.

This PR is mostly a port of the [print_id function](https://github.com/ruby/ruby/blob/e0c100ec96810b999c704fc73d60255484b63844/.gdbinit#L1073-L1113) from ruby's .gdbinit script, which takes a ruby method ID and returns the corresponding name as a string. The memory-walking logic is complex but has been very reliable in my testing so far. 

I wasn't able to locate the global symbol table for 2.6 or older, so support is currently limited to 2.7.x (and newer, hopefully). It may be possible to expand this with more investigation -- I probably missed something. I haven't tried rubies newer than 2.7.2 yet.

Here's a sample from tracing a simple puma server when it's under load:

```
Summary of profiling data so far:
% self  % total  name
 25.45    25.45  <c function> - sleep
 13.64    13.64  <c function> - expand_path
 11.82    49.09  <c function> - unknown
  3.64    20.91  block in <class:Base> - /Users/acj/.asdf/installs/ruby/2.7.1/lib/ruby/gems/2.7.0/gems/sinatra-2.1.0/lib/sinatra/base.rb
  3.64     4.55  <c function> - select
  3.64     3.64  <c function> - syswrite
  2.73     2.73  <c function> - write
  2.73     2.73  <c function> - peeraddr
  1.82    74.55  block in spawn_thread - /Users/acj/.asdf/installs/ruby/2.7.1/lib/ruby/gems/2.7.0/gems/puma-5.2.1/lib/puma/thread_pool.rb
  1.82    43.64  call - /Users/acj/.asdf/installs/ruby/2.7.1/lib/ruby/gems/2.7.0/gems/sinatra-2.1.0/lib/sinatra/base.rb
  1.82    41.82  call - /Users/acj/.asdf/installs/ruby/2.7.1/lib/ruby/gems/2.7.0/gems/rack-2.2.3/lib/rack/logger.rb
  1.82    30.91  call! - /Users/acj/.asdf/installs/ruby/2.7.1/lib/ruby/gems/2.7.0/gems/sinatra-2.1.0/lib/sinatra/base.rb
  1.82     2.73  append - /Users/acj/.asdf/installs/ruby/2.7.1/lib/ruby/gems/2.7.0/gems/puma-5.2.1/lib/puma/io_buffer.rb
  1.82     2.73  <c function> - each
[...]

PR #285, opened 2021-02-13
Add support for ruby 3.0.0

### LINKED ISSUE #278: Ruby 3 support
```bash
➜ sudo rbspy record --pid 15702
Press Ctrl+C to stop
thread '<unnamed>' panicked at 'Ruby version not supported yet: 3.0.0. Please create a GitHub issue and we'll fix it!', src/core/initialize.rs:353:14
```

PS: Thank you so much for continuing to support this!

acj:
I'm splitting this out of #282 as it may take a bit of spelunking. I've run bindgen and updated the usual places, but it doesn't currently work on 3.0.0:

Linux:
```
$ ruby -v
ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-linux]
$ ruby ci/ruby-programs/infinite.rb &
$ sudo env RUST_LOG=debug cargo run -- record --pid $(pgrep ruby)
Press Ctrl+C to stop
[2021-02-13T23:21:34Z DEBUG rbspy::core::address_finder::os_impl] symbol: Symbol: Value: 0x00317ac9 Size: 0x0006 Type: data object Bind: global Vis: default Section: 14 Name: ruby_version
[2021-02-13T23:21:34Z DEBUG rbspy::core::address_finder::os_impl] load header: Program Header: Type: LOAD Offset: 0x00030000 VirtAddr: 0x00030000 PhysAddr: 0x00030000 FileSize: 0x2a42a1 MemSize: 0x2a42a1 Flags: R E Align: 0x1000
[2021-02-13T23:21:34Z DEBUG rbspy::core::initialize] version: 3.0.0
[2021-02-13T23:21:34Z DEBUG rbspy::core::address_finder::os_impl] Trying to find address location another way
[2021-02-13T23:21:34Z DEBUG rbspy::core::address_finder::os_impl] bss_section header: SectionHeader { name: ".bss", shtype: 0x8, flags: 0x3, addr: 4095616, offset: 4091516, size: 67656, link: 0, info: 0, addralign: 32, entsize: 0 }
[2021-02-13T23:21:34Z DEBUG rbspy::core::address_finder::os_impl] read_addr: 7f9f675bde80
[2021-02-13T23:21:34Z DEBUG rbspy::core::address_finder::os_impl] successfully read data
Wrote raw data to /root/.cache/rbspy/records/rbspy-2021-02-13-KdNndnPMno.raw.gz
Writing formatted output to /root/.cache/rbspy/records/rbspy-2021-02-13-pxa4oxdRyE.flamegraph.svg
[2021-02-13T23:21:34Z ERROR inferno::flamegraph] No stack counts found
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Io(Custom { kind: InvalidData, error: "No stack counts found" })', src/ui/flamegraph.rs:38:89
```

macOS:
```
$ ruby -v
ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-darwin20]
$ ruby ci/ruby-programs/infinite.rb &
$ sudo env RUST_LOG=debug cargo run -- record --pid $(pgrep ruby)
Press Ctrl+C to stop
[2021-02-13T23:23:52Z DEBUG rbspy::core::initialize] version: 3.0.0
Wrote raw data to /Users/acj/.cache/rbspy/records/rbspy-2021-02-13-jqQNIRXiW6.raw.gz
Writing formatted output to /Users/acj/.cache/rbspy/records/rbspy-2021-02-13-cDlx202Und.flamegraph.svg
[2021-02-13T23:23:52Z ERROR inferno::flamegraph] No stack counts found
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Io(Custom { kind: InvalidData, error: "No stack counts found" })', src/ui/flamegraph.rs:38:9
```

If someone is interested in working on this, please feel free to jump in. Maybe it's something simple. I'm happy to help with testing but won't have time to investigate further until the C symbols work is done.

Resolves #278 (eventually)


acj:
Status update: There are significant thread-related changes in ruby 3.x due to the introduction of [ractors](https://github.com/ruby/ruby/blob/cfc23903df6bde6d79292d8a0291bc82e920b54e/doc/ractor.md) (ruby actors). This breaks several assumptions in rbspy when we try to look up the current execution context and use it to find stack frames. I've been slowly working through these issues and trying to fix them without making the version-specific code harder to reason about.

This code should currently work on Linux with ruby 2.7+. It builds but is broken (no stack frames) on macOS, which is next on my list to debug. I haven't tried Windows or FreeBSD yet.

acj:
It's able to find stack traces on macOS if I re-run bindgen on that platform, but then it's broken on Linux. A diff of the bindings turns up several changes but nothing directly related to our stack frame code (afaict). I wouldn't be surprised if there's a type size or alignment issue, though. I'll try to narrow it down, and if it can't be resolved then we may need to fork the bindings based on platform -- or find a way to avoid the problematic types.

acj:
This works on macOS now. 

I tried it on Windows tonight and didn't have much luck. The tests pass, but it seems like we're looking at the wrong memory locations when we try to get stack frames. Makes me wonder if we'll need to run bindgen separately on each platform to get the correct types (like I've already done for macOS and Linux). The main sticking point, afaict, is that the pthread primitives are platform-specific and differ pretty significantly in their size (e.g. struct vs union), so it's easy to jump to the wrong location in memory if the types and platform are mismatched. We need to step over a few of those pthread types as we're walking through the ractor and threads structs to get the execution context. If anyone knows of a shortcut to get the EC, please let me know.

acj:
> If anyone knows of a shortcut to get the EC

It's the opposite of a shortcut, but I found a way to get the EC address without needing to step over pthread types. The VM struct has a pointer to the current thread struct, which in turn has a pointer to the EC.

It works on Linux, macOS, and Windows without requiring separate bindings for each platform, which is a relief. rbspy seems to be generally broken on FreeBSD 12 for me (can't get stack frames, even on master with 2.7.2), so I'm keeping the build healthy but otherwise not spending time on it.

I'm going to clean up this branch a bit and then open it up for review (and maybe testing 🙏).

acj:
@jvns When you have time, this is ready for a look please. The logic in the `get_execution_context_from_vm` macro is a bit hacky, but it's worked well for me so far, and it sidesteps the maintenance headache of needing to run bindgen on all platforms each time we add a new version of ruby 3+. 

If we need to fall back to the bindgen approach (now or later), I think it would be good to invest some effort in automating all of it through GH Actions, maybe building on Igor's docker work, so that generating new bindings is as simple as running a CI job. (As I'm writing this, it seems like a good thing to do regardless, hah. :)

src/core/ruby_version.rs
```diff
@@ -181,6 +188,78 @@ macro_rules! ruby_version_v2_7_x(
...
+            // require us to maintain separate ruby-structs bindings for each platform due to
+            // their varying sizes and alignments.
+            let vm_addr: usize = source.copy_struct(ruby_vm_address_location)
+                .context(ruby_vm_address_location)?;
+            let vm: rb_vm_struct = source.copy_struct(vm_addr as usize)
+                .context(vm_addr)?;
+
+            // Seek forward in the ractor struct, looking for the main thread's address. The 
+            // execution context pointer is in the memory word just before it 
+            // (see rb_ractor_struct). The initial offset was found through experiment.
```
jvns, line 242: is the idea that [this pointer](https://github.com/ruby/ruby/blob/master/ractor_core.h#L99) the execution context pointer and the size of [these struct members](https://github.com/ruby/ruby/blob/master/ractor_core.h#L83-L91) is about 520?

src/core/ruby_version.rs
```diff
@@ -181,6 +188,78 @@ macro_rules! ruby_version_v2_7_x(
...
+            let initial_offset = 520;
+            for i in initial_offset..(initial_offset+500) {
+                let candidate_main_thread_address: usize = source.copy_struct(vm.ractor.main_ractor as usize + i)
```
jvns, line 245: Would it make sense here to do 1 big copy here (into an array of addresses) instead of many small copies? It would be faster but it only happens once.

jvns:
lgtm! This seems very reasonable to me and it feels less hacky than some of the other things we do in rbspy like `is_maybe_thread` (this follows way less pointers!). Thanks for the comments too :)

I'm so happy you did this!

src/core/ruby_version.rs
acj: Yep, exactly. I've found that there are usually a few pointers to the EC in those early structs as well (maybe in the linked lists or something), so it's crucial to jump forward a bit instead of starting at the top of the ractor struct. I'll add a note about this before we merge.

src/core/ruby_version.rs
acj: I like that. Getting the memory addresses into a standard vector so that it can be iterated over will probably make this logic easier to follow, too.

acj:
This is in good shape now, I think. The Travis build gets through the tests and then fails when it uses libjemalloc, but I'm able to use libjemalloc and libtcmalloc locally with ruby 2.4. The malloc tests have been flaky in CI for me before.

If anyone has time to help test this, please do.

I'll probably merge tomorrow if there are no objections, and then cut a release next week.

PR #290, opened 2021-03-07
Fix flaky subprocess test

acj:
(... hopefully.)

The `test_spawn_record_children_subprocesses` has been failing consistently for me locally (macOS 11.2.1, 2019 MBP). I noticed that the test process was returning an error immediately -- before ruby_forks.rb had exited, and maybe before it had even started -- which made me wonder whether there was a startup delay that we weren't accounting for. Indeed, it seems to take a few hundred milliseconds before we can get a handle on the process. If we wait for that handle before spawning the recorders, it works consistently for me.

After fixing that, I noticed an intermittent related issue where the trace sender channel was filling up. I think that's the reason for the timeouts we've seen in CI recently -- the sender channel fills up, which blocks some recorders from exiting, which in turn blocks the subprocess test from collecting the results.

There is similar retry logic elsewhere (e.g. get_ruby_version_retry) that may explain why we don't see this behavior outside of the tests.

cc @igorwwwwwwwwwwwwwwwwwwww -- thanks for your earlier work on this 👍

### DIFF
--- a/ci/ruby-programs/ruby_forks.rb
+++ b/ci/ruby-programs/ruby_forks.rb
@@ -1,4 +1,4 @@
-SLEEP_TIME = 2
+SLEEP_TIME = 5
 
 suprocess_cmd = <<-CMD
   pid1_1 = spawn(ENV, RbConfig.ruby, "-esleep(#{SLEEP_TIME})")
--- a/src/core/initialize.rs
+++ b/src/core/initialize.rs
@@ -2,7 +2,7 @@ use crate::core::address_finder::*;
 use crate::core::address_finder;
 use proc_maps::MapRange;
 use crate::core::ruby_version;
-use crate::core::types::{MemoryCopyError, Pid, Process,
+use crate::core::types::{MemoryCopyError, Pid, Process, ProcessRetry,
                          ProcessMemory, StackTrace};
 
 use failure::Error;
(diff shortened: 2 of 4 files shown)


jvns:
Amazing! :heart:

jvns:
this is amazing! :heart: I was going to merge this but I noticed that it's a draft -- feel free to merge it whenever you'd like :)

acj:
@jvns thanks! I realized that the problem ran a bit deeper and put it through another cycle. I'll merge this if the tests seem stable after the most recent changes.

> There is similar retry logic elsewhere (e.g. get_ruby_version_retry) that may explain why we don't see this behavior outside of the tests.

After I wrote that, I remembered that I _have_ seen this behavior when trying to profile a ruby CLI tool locally. The most recent changes move the retry logic into `initialize` and the recorder itself, and that seems to fix both the subprocess test and my CLI tool.

I also tweaked the subprocess test code to kill the ruby processes as soon as we have the expected data. It's a slight hack, but I think it'll make the tests more reliable and also a little faster in the best case.

acj:
Killing the ruby processes early caused some flakiness on Windows, so I took that out. Letting ruby_forks.rb run a little longer (~5s) should work okay for now. I'll try to speed it up later.

acj:
I'm still seeing timeouts in the subprocess test occasionally. Hopefully they'll be less common now, but it seems like there's more tuning to do.

PR #291, opened 2021-03-13
Fix special variable handling when resolving C functions

### LINKED ISSUE #289: Crash: thread '<unknown>' has overflowed its stack
I saw this while running the latest rbspy `v0.3.13`.

Command:

```
sudo rbspy record -p $(pgrep -n -f 'puma ') --subprocesses --duration 10
```

Crash:

```
thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
```

gdb on the core dump gave me:

```
(gdb) bt
#0  0x000000000070fe1f in __restore_sigs ()
#1  0x000000000070fe60 in raise ()
#2  0x00007f821ffdc4b0 in ?? ()
#3  0x00007f82203edac8 in ?? ()
#4  0x00000000006e0242 in std::io::Write::write_fmt::h9ff815313448b677 () at library/std/src/io/mod.rs:1518
#5  0x00007f8449189af0 in ?? ()
#6  0x00007f82203edac8 in ?? ()
#7  0x000000000070ddb2 in abort ()
#8  0x0000000000a3df28 in ?? ()
#9  0x0000000000000002 in ?? ()
#10 0x0000000000000000 in ?? ()
```

It's happening consistently. Could be related to something new.

acj:
Fixes #289. This crash seems to be fairly common on busy production systems but is hard to reproduce otherwise. I've tried a few small programs that use [special variables](https://ruby.fandom.com/wiki/Special_variable) in different ways without seeing them show up in a stack frame. So, no unit test for now.

I've run this for over an hour on a busy service without any issue, whereas it would crash reliably within ~5 minutes before.


jvns:
lgtm!

PR #295, opened 2021-03-22
Update the remoteprocess/proc-maps crates

benfred:
The latest remoteprocess code includes a workaround for the case where the linux kernel
was compiled without process_vm_readv support (by falling back to /proc/PID/mem:
https://github.com/benfred/read-process-memory/pull/4).

Closes #196

Issue #263, opened 2020-06-05
spontaneous exits?


doriantaylor:
I've seen a few bug reports about `rbspy` exiting spontaneously. I'm adding to these with a report of `rbspy record` exiting on its own after a few seconds (with `$? == 0`); unlike the other reports there's no warning or error message. I'm using a version of `rbspy` I just compiled from `master` against `ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-darwin19]` from homebrew on macOS 10.15.5.

jvns:
Do you have any ideas for how to reproduce? Unfortunately without steps to reproduce I can't do much.

If you're interested in hacking on the problem and trying to track down what's going on in the code yourself I'd really love the help -- I bet it's possible to make some progress with just some print statements. happy to give pointers about where to start or answer questions too -- there's a description of rbspy's architecture in the repo as well that might be helpful.

doriantaylor:
Other than running it against some proprietary employer code I don't know how I would; I can try to create something standalone and see if it reproduces but that'll be a few weeks before I have an opening for something like that.

It (ostensibly) doesn't crash or anything, it just…quits. I'm assuming it isn't normal operation?

(aside: I'm trying `rbspy` because`ruby-prof` was crashing with `Illegal instruction: 4`; trying to smoke out [my recently-inherited lmdb bindings](/doriantaylor/lmdb))

Issue #243, opened 2019-12-16
Recording stops prematurely with no errors


wagenet:
I'v been consistently finding that my recordings stop prematurely (even with a duration specified) without any sort of notification as to why. I checked the docs and couldn't find any information about what might cause this.

wagenet:
This seems to be specific to macOS.

wagenet:
I realized I could turn on logger with RUST_LOG and discovered that it is stopping because of `MemoryCopyError::ProcessEnded`.

acj:
Possibly related to #263 even though the error output is different. I've run into this as well on busy systems, and it's happened a couple of times while I've been working on ruby 3 support and profiling simple sinatra servers. The latter makes me think that the early exits happen because of an illegal memory access or something similar. I'm not sure why we'd get a `process exited` error in that case, though. Maybe a library bug where macOS's error code is misinterpreted?

I'll try to chase this down once the ruby 3 work lands.

PR #296, opened 2021-03-24
Synchronous subprocess test

acj:
We've still been seeing subprocess test failures in CI, especially on Windows. This PR refactors the subprocess test to coordinate with ruby_forks.rb using the filesystem instead of depending on the ruby processes' startup speed. The last commit is an ugly workaround for process exit behavior that hopefully can be fixed (maybe by #295).

### DIFF
--- a/ci/ruby-programs/ruby_forks.rb
+++ b/ci/ruby-programs/ruby_forks.rb
@@ -1,17 +1,23 @@
-SLEEP_TIME = 5
+coordination_path = ARGV[0].gsub('\\', '/')
 
-suprocess_cmd = <<-CMD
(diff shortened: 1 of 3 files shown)


acj:
I can't see the build output in Actions anymore, but one of the previous builds got all `Ok` results from the subprocesses but somehow saw 3 PIDs instead of 4. My best guess is that we ran into an edge case like #243 or #263 and one of the ruby processes exited before we could get a stack trace. If that's the case, this PR won't fix the whole problem, but it'll make CI less flaky (I hope) and let us focus on fixing real bugs in the stack trace fetching code.

PR #310, opened 2021-04-11
Add workflow for generating ruby bindings

acj:
It's currently a tedious manual process to generate bindings for new (or existing) ruby versions. The process is predictable, though, and I wondered if it could be fully automated. Happily, the answer is yes.

This PR adds a new GitHub Actions workflow that generates the bindings for (almost) all of our supported rubies from 1.9.3 through 3.0.1. The workflow currently needs to be triggered manually in the Actions tab. Assuming the builds succeed, it will gather the generated source files and create a new branch that contains any changes or additions. We could eventually have it open a PR if we wanted.

The bindgen script can still be run in isolation just as before, but I'm hoping that we won't need to do that very often.

If we ever need to generate separate bindings for each platform (like I almost needed to for ruby 3.x), this should make the process a lot more scalable.

Samples:
- [Workflow](https://github.com/acj/rbspy/actions/runs/738911100)
- [Branch diff](https://github.com/rbspy/rbspy/compare/master...acj:generate-ruby-bindings-38)

cc @rbspy/maintainers @daniellockyer @igorwwwwwwwwwwwwwwwwwwww @liaden

### DIFF
--- a/.github/workflows/ruby-bindings.yml
+++ b/.github/workflows/ruby-bindings.yml
@@ -1,0 +1,119 @@
+name: ruby-bindings
+on:
+  workflow_dispatch:
+jobs:
+  generate:
+    name: Generate Ruby bindings
+    env:
+      CARGO: cargo
+      TARGET_FLAGS: --target ${{ matrix.target }}
+      RUST_BACKTRACE: 1
+    timeout-minutes: 20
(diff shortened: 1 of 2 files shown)


igorwwwwwwwwwwwwwwwwwwww:
LGTM!

PR #311, opened 2021-04-13
Fix subprocess exit bug

### LINKED ISSUE #287: Mac CI build is flaky
The MacOS CI build has a race and fails a lot. I'm not sure why the flaky test doesn't have the same flakiness issue on Linux.

acj:
rbspy currently stops profiling when any one of the processes it's tracking exits. This creates a subtle race condition for the subprocess test, which has been very flaky on macOS and Windows, and also seems like undesirable behavior since subprocesses are expected to come and go in many applications. There may also be use cases where the root process is expected to exit before its subprocesses, but I think this change will cover the 99% case pretty well.

I realized this while experimenting with `--test-threads` for an unrelated issue. When I limited the threads to one or two (which is close to CI worker specs), the subprocess test consistently failed.

@jvns, when you have a moment could you please take a peek at this? I'm curious if there's a historical reason why we were exiting as soon as the first process ended.

Fixes #287

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -501,7 +501,6 @@ fn spawn_recorder_children(
 }
 
 // TODO: Find a more reliable way to test this on Windows hosts
-#[cfg(not(target_os = "windows"))]
 #[test]
 fn test_spawn_record_children_subprocesses() {
     let which = if cfg!(target_os = "windows") {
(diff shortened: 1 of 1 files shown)


acj:
I feel good about this, so going to merge and include it in v0.5. Let me know if there are any concerns.

jvns:
this looks good to me, I don't think there was a specific reason it worked this way

acj:
Okay, cool. Thanks for looking! Hopefully this bug is gone for good now.

PR #316, opened 2021-04-18
Pause target process when retrieving stack frames

### LINKED ISSUE #153: Support pausing the process while getting the stack
Could probably do this with SIGSTOP/SIGCONT

### LINKED ISSUE #315: Inconsistent stack samples
Hello I'm not an expert in `rbspy` so I hope I'm understanding this issue correctly and it is in fact a bug.

```ruby
def foo
  bar
end

def bar
  baz
  bob
end

def baz
  [1, 2, 3].map(&:succ)
end

def bob
  [1, 2, 3]
end

loop do
  foo
end
```

When I profile this program using `rbspy` `0.5.0` with both Ruby 2.7.2 and 3.0.1 I get a flame graph that shows for example `succ` being called from `bob`, as well as from `baz`.

Screenshot 2021-04-15 at 16 22 16

The raw logs also show it:

```
{
   "trace":[
      {
         "name":"succ [c function]",
         "relative_path":"(unknown)",
         "absolute_path":null,
         "lineno":0
      },
      {
         "name":"map [c function]",
         "relative_path":"(unknown)",
         "absolute_path":null,
         "lineno":0
      },
      {
         "name":"bob",
         "relative_path":"tool/diffprof/rbspy-query.rb",
         "absolute_path":"/Users/chrisseaton/Documents/truffleruby/tool/diffprof/rbspy-query.rb",
         "lineno":16
      },
      {
         "name":"bar",
         "relative_path":"tool/diffprof/rbspy-query.rb",
         "absolute_path":"/Users/chrisseaton/Documents/truffleruby/tool/diffprof/rbspy-query.rb",
         "lineno":8
      },
      {
         "name":"foo",
         "relative_path":"tool/diffprof/rbspy-query.rb",
         "absolute_path":"/Users/chrisseaton/Documents/truffleruby/tool/diffprof/rbspy-query.rb",
         "lineno":3
      },
      {
         "name":"block in <main>",
         "relative_path":"tool/diffprof/rbspy-query.rb",
         "absolute_path":"/Users/chrisseaton/Documents/truffleruby/tool/diffprof/rbspy-query.rb",
         "lineno":20
      },
      {
         "name":"loop [c function]",
         "relative_path":"(unknown)",
         "absolute_path":null,
         "lineno":0
      },
      {
         "name":"<main>",
         "relative_path":"tool/diffprof/rbspy-query.rb",
         "absolute_path":"/Users/chrisseaton/Documents/truffleruby/tool/diffprof/rbspy-query.rb",
         "lineno":18
      }
   ],
   "pid":8126,
   "thread_id":4547739136,
   "time":{
      "secs_since_epoch":1618497446,
      "nanos_since_epoch":257731000
   }
}
```

I don't understand how it's getting this stack trace sample from this program?

As I say I'm not an expert, but is it possible that this stack trace really ends at the call to `bob`, and the two entries on top of that are memory left on the call stack which hasn't been cleared, or the stack pointer is incorrectly sampled?

Possibly unrelated, but there also seem to be incorrect line numbers - it's like each activation always points to the `end` of the method, even if for example it's calling `baz` from `bar`, so not at the end of the method.

Thanks in advance for any help!

acj:
We don't currently pause the target process when calling `get_stack_frame`, which results in incorrect and misleading profiler output like we see in #315. The `remoteprocess` crate has support for pausing processes on each platform. The output seems much more accurate in my testing, so I'm going to enable it.

Using higher sampling frequencies will degrade performance of the target process while rbspy is running, but 1) this is already true, at least on some systems, because of higher CPU usage, and 2) it feels like a reasonable tradeoff for the sake of correctness. If we find that it's a problem, we could consider adding a flag to disable pausing.

Fixes #153. Fixes #315.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -827,7 +827,7 @@ dependencies = [
  "libc",
  "libproc",
  "log",
- "nix 0.20.0",
+ "nix 0.19.1",
  "proc-maps",
  "rand 0.8.3",
  "rbspy-ruby-structs",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ tempdir = "0.3"
 thiserror = "1.0.24"
 
 [target.'cfg(unix)'.dependencies]
-nix = "0.20.0"
+nix = "0.19.1"
 
 [target.'cfg(target_os="macos")'.dependencies]
 libproc = "0.9.1"
--- a/src/core/initialize.rs
+++ b/src/core/initialize.rs
@@ -12,7 +12,7 @@ use std::time::Duration;
 /**
  * Initialization code for the profiler.
  *
- * The only public function here is `initialize(pid: Pid)`, which returns a struct which you can
+ * The only public function here is `initialize`, which returns a struct which you can
  * call `get_trace()` on to get a stack trace.
  *
  * Core responsibilities of this code:
@@ -21,7 +21,7 @@ use std::time::Duration;
  *   * Find the right stack trace function for the Ruby version we found
  *   * Package all that up into a struct that the user can use to get stack traces.
  */
-pub fn initialize(pid: Pid) -> Result<StackTraceGetter> {
+pub fn initialize(pid: Pid, lock_process: bool) -> Result<StackTraceGetter> {
     let process = Process::new_with_retry(pid)?;
     let (
         current_thread_addr_location,
@@ -37,6 +37,7 @@ pub fn initialize(pid: Pid) -> Result<StackTraceGetter> {
         global_symbols_addr_location,
         stack_trace_function,
         reinit_count: 0,
+        lock_process,
     })
 }
 
@@ -48,6 +49,7 @@ pub struct StackTraceGetter {
     global_symbols_addr_location: Option<usize>,
     stack_trace_function: StackTraceFn,
     reinit_count: u32,
+    lock_process: bool,
 }
 
 impl StackTraceGetter {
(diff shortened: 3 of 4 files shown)


jvns:
It looks like this is py-spy's approach -- pause by default and have a `--nonblocking` flag to disable pausing https://github.com/benfred/py-spy#how-can-i-avoid-pausing-the-python-program.

benfred:
I did some testing on locking with rbspy when I originally put the locking code into py-spy - my takeaway was that locking the process isn't as important with rbspy as it is with py-spy, because of how stacks are represented internally on the python/ruby interpreters.

I'll write up a quick blog post on this in the next couple days going into some more details on this, but the big takeaway is that the errors from not locking are *much* worse with py-spy than they are with rbspy. 

This locking code obviously helps in some situations (like #315) - so is good to add. However, I think we should add a commandline flag to turn off/on the locking code, and that the default for rbspy maybe should be to run without locking -  like the opposite of py-spy which defaults to locking. (I think the fact that this hasn't been raised as an issue in several years for rbspy but was flagged with py-spy in its first 2 months speaks to the relative importance of locking with rbspy/py-spy).

acj:
Thanks for the context @benfred. I like the `--nonblocking` flag and am happy to add it before this is merged.

I'm pretty firmly on the side of doing the safe and correct thing by default -- it's what I expect from this type of tool. If someone wants to knowingly turn off a safety feature to improve performance, that's fine, but IMO it should be opt-out.

It's a fair point that we haven't gotten many complaints related to this type of thing. I think the stack correctness issues may have been obscured by other bugs that have been fixed recently, and folks may have used rbspy to get a rough initial guess at where the slow spots are before switching to something more invasive (and, currently, more accurate) like stackprof. If they felt more confident in rbspy's output, it might make a difference in their chosen workflow. That was the case for me when I started working with it -- it was incredible but a bit rough around the edges, so I wanted a second tool's opinion -- and is definitely influencing my feelings here.

benfred:
I finally got around to writing that post on the nonblocking mode for rbspy vs py-spy :
https://www.benfrederickson.com/why-python-needs-paused-during-profiling/

PR #327, opened 2021-06-07
Add ability to collect traces only when the CPU is active (--on-cpu)

michelhe:
This change adds on what #127 tried to bring to allow measuring how much time is spent on CPU, (but implemented slightly different and more up to date).

We filter traces that are off-cpu using 2 checks:
- OS-specific, currently using the `remoteprocess` crate to determine if there are any active threads at this point in time.
- Using the `rb_thread_struct.status` field to figure out if a thread is in `THREAD_RUNNABLE` or not when we come to get a trace from it. The problem is that this is only a heuristic and will be inaccurate in cases when the process gets preempted by the OS while it has runnable threads, so we rely on OS-specific checks as well when possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,13 +16,20 @@ production**.
 `rbspy` lets you record profiling data, save the raw profiling data to disk, and then analyze it in
 a variety of different ways later on.
 
-## only wall-clock profiling
+## profiling options
 
 There are 2 main ways to profile code -- you can either profile everything the
 application does (including waiting), or only profile when the application is using the CPU.
 
-rbspy profiles everything the program does (including waiting) -- there's no
-option to just profile when the program is using the CPU.
+By default, rbspy profiles everything the program does (including waiting).
+To sample only when the program is using the CPU, use the `--on-cpu` flag.
+
+`--on-cpu` is implemented by 2 layers of filtering using the following sources of information:
+- OS level thread information (only available for linux, macos and windows)
+- current thread state in the ruby interpreter.
+
+In addition, rbspy can either profile in a blocking or nonblocking mode.
+The latter is less accurate but does not require rbspy to stop the live process to take samples.
 
 ## Documentation
 
--- a/src/core/initialize.rs
+++ b/src/core/initialize.rs
@@ -26,6 +26,7 @@ pub fn initialize(
     pid: Pid,
     lock_process: bool,
     force_version: Option<String>,
+    on_cpu: bool
 ) -> Result<StackTraceGetter> {
     #[cfg(all(windows, target_arch = "x86_64"))]
     if is_wow64_process(pid).context("check wow64 process")? {
@@ -51,6 +52,7 @@ pub fn initialize(
         reinit_count: 0,
         lock_process,
         force_version,
+        on_cpu,
     })
 }
 
@@ -64,21 +66,29 @@ pub struct StackTraceGetter {
     reinit_count: u32,
     lock_process: bool,
     force_version: Option<String>,
+    on_cpu: bool,
 }
 
 impl StackTraceGetter {
-    pub fn get_trace(&mut self) -> Result<StackTrace> {
+    pub fn get_trace(&mut self) -> Result<Option<StackTrace>> {
+        /* First, trying OS specific checks to determine whether the process is on CPU or not.
+         * This comes before locking the process because in most operating systems locking
+         * means the process is stopped */
+        if self.on_cpu && !self.is_on_cpu_os_specific()? {
+            return Ok(None);
+        }
         match self.get_trace_from_current_thread() {
-            Ok(mut trace) => {
+            Ok(Some(mut trace)) => {
                 return {
                     /* This is a spike to enrich the trace with the pid.
                      * This is needed, because remoteprocess' ProcessMemory
                      * trait does not expose pid.
                      */
                     trace.pid = Some(self.process.pid);
-                    Ok(trace)
+                    Ok(Some(trace))
                 };
             }
+            Ok(None) => return Ok(None),
             Err(MemoryCopyError::InvalidAddressError(addr))
                 if addr == self.current_thread_addr_location => {}
             Err(e) => {
@@ -95,7 +105,25 @@ impl StackTraceGetter {
             .context("get trace from current thread")?)
     }
 
-    fn get_trace_from_current_thread(&self) -> Result<StackTrace, MemoryCopyError> {
+    #[cfg(any(target_os = "macos", target_os = "windows", target_os = "linux"))]
+    fn is_on_cpu_os_specific(&self) -> Result<bool> {
+        // remoteprocess crate exposes a Thread.active() method for each of these targets
+        for thread in self.process.threads()?.iter() {
+            if thread.active()? {
+                return Ok(true);
+            }
+        }
+        Ok(false)
+    }
+
+    #[cfg(not(any(target_os = "macos", target_os = "windows", target_os = "linux")))]
+    fn is_on_cpu_os_specific(&self) -> Result<bool> {
+        /* We don't have OS specific checks for these targets,
+         * so fallback to using the interpreter based check down the line */
+        Ok(true)
+    }
+
+    fn get_trace_from_current_thread(&self) -> Result<Option<StackTrace>, MemoryCopyError> {
         let stack_trace_function = &self.stack_trace_function;
 
         let _lock;
@@ -112,6 +140,7 @@ impl StackTraceGetter {
             self.global_symbols_addr_location,
             &self.process,
             self.process.pid,
+            self.on_cpu,
         )
     }
 
@@ -140,8 +169,16 @@ pub type IsMaybeThreadFn = Box<dyn Fn(usize, usize, &Process, &[MapRange]) -> bo
 
 // Everything below here is private
 
-type StackTraceFn =
-    Box<dyn Fn(usize, usize, Option<usize>, &Process, Pid) -> Result<StackTrace, MemoryCopyError>>;
+type StackTraceFn = Box<
+    dyn Fn(
+        usize,
+        usize,
+        Option<usize>,
+        &Process,
+        Pid,
+        bool,
+    ) -> Result<Option<StackTrace>, MemoryCopyError>,
+>;
 
 fn get_process_ruby_state(
     pid: Pid,
@@ -564,17 +601,39 @@ mod tests {
         result.expect("unexpected error");
     }
 
-    #[test]
-    #[cfg(target_os = "linux")]
-    fn test_get_trace() {
-        // Test getting a stack trace from a real running program using system Ruby
-        let cmd = RubyScript::new("./ci/ruby-programs/infinite.rb");
-        let pid = cmd.id() as Pid;
-        let mut getter = initialize(pid, true, None).unwrap();
-        std::thread::sleep(std::time::Duration::from_millis(50));
-        let trace = getter.get_trace();
-        assert!(trace.is_ok());
-        assert_eq!(trace.unwrap().pid, Some(pid));
+    macro_rules! generate_get_trace_tests {
+
+        ($($name:ident($lock:expr, $on_cpu:expr),)*) => {
+        $(
+            #[test]
+            #[cfg(target_os = "linux")]
+            fn $name() {
+
+                // Test getting a stack trace from a real running program using system Ruby
+                let cmd = if $on_cpu {
+                    RubyScript::new("./ci/ruby-programs/off_cpu.rb");
+                } else {
+                    RubyScript::new("./ci/ruby-programs/infinite.rb");
+                };
+                let pid = cmd.id() as Pid;
+                let mut getter = initialize(pid, $lock, None, $on_cpu).unwrap();
+                std::thread::sleep(std::time::Duration::from_millis(200));
+                let trace = getter.get_trace();
+                assert!(trace.is_ok());
+                let trace = trace.unwrap();
+                if $on_cpu {
+                    assert!(trace.is_none());
+                } else {
+                    assert_eq!(trace.pid, Some(pid));
+                }
+            }
+        )*
+        }
+    }
+
+    generate_get_trace_tests! {
+        test_get_trace(true, false),
+        test_get_trace_on_cpu(true, true),
     }
 
     #[test]
@@ -590,7 +649,7 @@ mod tests {
             .spawn()
             .unwrap();
         let pid = cmd.id() as Pid;
-        let mut getter = initialize(pid, true, None).expect("initialize");
+        let mut getter = initialize(pid, true, None, false).expect("initialize");
 
         std::thread::sleep(std::time::Duration::from_millis(50));
         let trace1 = getter.get_trace();
@@ -638,7 +697,7 @@ mod tests {
         }
 
         let mut cmd = RubyScript::new("ci/ruby-programs/infinite.rb");
-        let mut getter = crate::core::initialize::initialize(cmd.id(), true, None).unwrap();
+        let mut getter = crate::core::initialize::initialize(cmd.id(), true, None, false).unwrap();
 
         cmd.kill().expect("couldn't clean up test process");
 
--- a/src/core/ruby_version.rs
+++ b/src/core/ruby_version.rs
@@ -25,6 +25,7 @@ macro_rules! ruby_version_v_1_9_1(
             get_lineno_1_9_0!();
             get_stack_frame_1_9_1!();
             stack_field_1_9_0!();
+            get_thread_status_1_9_0!();
             get_thread_id_1_9_0!();
             get_cfunc_name_unsupported!();
         }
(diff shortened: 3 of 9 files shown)


acj:
Hey, thanks for this. We've been wanting on/off CPU filtering for a long time. I'll try to do a deep review sometime in the next week or so.

README.md
```diff
@@ -16,14 +16,20 @@ production**.
...
-## only wall-clock profiling
+## profiling options
 
 There are 2 main ways to profile code -- you can either profile everything the
 application does (including waiting), or only profile when the application is using the CPU.
 
-rbspy profiles everything the program does (including waiting) -- there's no
-option to just profile when the program is using the CPU.
+By defeault, rbspy profiles everything the program does (including waiting).
+There is an experimantal option to profile only when the program is using the CPU. (`--on-cpu`)
```
Jongy, line 25:
```suggestion
There is an experimental option to profile only when the program is using the CPU. (`--on-cpu`)
```

src/core/initialize.rs
```diff
@@ -74,7 +83,25 @@ impl StackTraceGetter {
...
+            }
+        }
+        Ok(false)
+    }
+
+    #[cfg(not(any(target_os = "macos", target_os = "windows", target_os = "linux")))]
+    fn is_on_cpu_os_specific(&self) -> Result<bool> {
+        /* We don't have OS specific checks for these targets,
+         * so fallback to using the interpreter based check down the line */
+        Ok(false)
```
Jongy, line 101: If you return `false` here and `on_cpu` is used, then nothing will be collected at all.

In this configuration - perhaps just deny using `on_cpu`?

src/core/initialize.rs
```diff
@@ -74,7 +83,25 @@ impl StackTraceGetter {
         Ok(self.get_trace_from_current_thread()?)
     }
 
-    fn get_trace_from_current_thread(&self) -> Result<StackTrace, MemoryCopyError> {
+    #[cfg(any(target_os = "macos", target_os = "windows", target_os = "linux"))]
+    fn is_on_cpu_os_specific(&self) -> Result<bool> {
+        // remoteprocess crate exposes a Thread.active() method for each of these targets
+        for thread in self.process.threads()?.iter() {
```
Jongy, line 89: This is not the logic you want - you need to check the "current" thread that we're going to get stacktraces from, not just any thread in the process

src/core/ruby_version.rs
```diff
@@ -277,17 +285,29 @@ macro_rules! get_stack_trace(
             ruby_global_symbols_address_location: Option<usize>,
             source: &T,
             pid: Pid,
-        ) -> Result<StackTrace, MemoryCopyError> {
+            on_cpu: bool,
+        ) -> Result<Option<StackTrace>, MemoryCopyError> {
             let thread: $thread_type = get_execution_context(ruby_current_thread_address_location, ruby_vm_address_location, source)
                 .context(ruby_current_thread_address_location)?;
 
+            let thread_status = get_thread_status(&thread, source)?;
```
Jongy, line 293: Tiny optimization - these 2 calls require reading into the remote's memory, so you can avoid them unless actually needed (don't need the status if not `on_cpu`, and don't need `thread_id` if the `on_cpu` check doesn't pass)

Jongy (commented):
Some comments for @michelhe :)

src/core/initialize.rs
michelhe: Nice catch, meant to return true 😅

src/core/initialize.rs
michelhe: We don't know which OS thread the "current" ruby thread until we start examining the interpreter's memory, so the heuristic is: "if any thread in the thread group is on CPU"

Jongy: If the "current" ruby thread is not active OS-wise, it's surely not active ruby-wise (to be precise: it may be marked active ruby-wise, this just means it's preempted). Another thread may be running, doing non-ruby stuff, and the thread doing ruby-stuff may be preempted, and you will incorrectly account that as the ruby code)

src/core/ruby_version.rs
```diff
@@ -277,17 +285,28 @@ macro_rules! get_stack_trace(
...
             source: &T,
             pid: Pid,
-        ) -> Result<StackTrace, MemoryCopyError> {
+            on_cpu: bool,
+        ) -> Result<Option<StackTrace>, MemoryCopyError> {
             let thread: $thread_type = get_execution_context(ruby_current_thread_address_location, ruby_vm_address_location, source)
                 .context(ruby_current_thread_address_location)?;
 
+            // testing the thread state in the interpreter.
+            if on_cpu && get_thread_status(&thread, source)? != 0 /* THREAD_RUNNABLE */ {
```
Jongy, line 294: I don't know ruby well enough - is this the equivalent of CPython's GIL? If so, it's probably better to add this as another option (compare to py-spy's `--active` for on-CPU, and `--gil` for GIL; 2 different options that can go together, but I might want `--active` only w/o `--gil`)

acj: My understanding is that they're very similar. Only one thread per VM can be executing ruby code at a time. Native C extensions can explicitly release the GVL, though, so that ruby and C code can execute in parallel when it's safe to do so. The situation is changing a bit in ruby 3 with the addition of ractors, which have their own lock and will probably need different logic for on/off CPU analysis. rbspy only has basic support for ractors right now (only looks at the main one), so we may need to defer that work until later.

I'm not opposed to introducing a `--gvl` option if we can find a reliable way to detect which thread holds the lock, but I'd recommend doing that in a future PR so that we can keep this one small-ish and focused. We might also want an `--off-cpu` option. Lots of future paths to explore here.

It seems that this `get_thread_status` check is the key mechanism, and the earlier check using remoteprocess's `Thread::active` is a fast/cheap filter to skip `get_trace` if we know that none of the process's threads is active (from the OS's perspective). Is that accurate?

acj, line 294: It should be possible to use `rb_thread_status_THREAD_RUNNABLE` here.

src/core/initialize.rs
acj: I touched on this in another comment, but if this is meant as an initial filter and the more accurate check happens later (via `get_thread_status`), then I think it's okay. Unless the remoteprocess functions we're calling here are expensive, this should reduce rbspy's CPU usage a bit when we're doing on-CPU profiling. It might be worth it to check that assumption, though -- I don't know offhand how expensive this loop is for a typical program.

README.md
```diff
@@ -16,14 +16,20 @@ production**.
...
 
-## only wall-clock profiling
+## profiling options
 
 There are 2 main ways to profile code -- you can either profile everything the
 application does (including waiting), or only profile when the application is using the CPU.
 
-rbspy profiles everything the program does (including waiting) -- there's no
-option to just profile when the program is using the CPU.
+By defeault, rbspy profiles everything the program does (including waiting).
```
acj, line 24:
```suggestion
By default, rbspy profiles everything the program does (including waiting).
```

README.md
```diff
@@ -16,14 +16,20 @@ production**.
...
+There is an experimental option to profile only when the program is using the CPU. (`--on-cpu`)
```
acj, line 25: > experimental option

Are there any concerns or caveats that we should document?  What's blocking this from being a non-experimental feature?

src/core/ruby_version.rs
Jongy: Thanks for the brief Ruby internals :)

> I'm not opposed to introducing a --gvl option if we can find a reliable way to detect which thread holds the lock,

I might have missed that - isn't it what we expect `THREAD_RUNNABLE` to tell? Or does runnable mean "can be run given that it obtains the GVL"? If the latter, then I think that looking at it doesn't make any sense at all.

> We might also want an --off-cpu option. Lots of future paths to explore here.

Yup. Well, basically, no `--on-cpu` means `--off-cpu` now :)

> It seems that this get_thread_status check is the key mechanism, and the earlier check using remoteprocess's Thread::active is a fast/cheap filter to skip get_trace if we know that none of the process's threads is active (from the OS's perspective). Is that accurate?

To my understanding - not precisely, I'll give an elaborated answer on the other discussion :sweat_smile:

src/core/initialize.rs
Jongy: @acj this is just not accurate enough - imagine this case: 1 thread in the application is running some non-related C code doing whatever (and it's active, so `thread.active()` returns `true`). Meanwhile, another thread is running Ruby code (so it has `rb_thread_status_THREAD_RUNNABLE`, I guess?). But the OS decided that it's time that the Ruby thread relinquishes CPU, and it gets preempted. Its `rb_thread_status` remains as it was.

So, per the 2 checks:
* the "process" is active, because one of its threads is active.
* the "Ruby thread" is also "active" because its status is `THREAD_RUNNABLE`.

And the result is that we collect Ruby stacks from a thread that's not currently running!

About the cost of `thread.active()` - it's not very expensive. It's a single read from `/proc/tid/stat`, so 3 syscalls (`open`, `read`, `close`). Comparing to `get_thread_status` - which is only one `process_vm_readv` IIRC, so faster; but it should be benched & compared to reading `/proc/tid/stat` if we want a real comparison.

acj: Yeah, sorry, I didn't mean to suggest that the entire approach is okay. I understood your first example and agree that it's something we need to solve (I'm still thinking about this). What I meant to say is that this `for` loop seems reasonable, provided that we can somehow do an accurate check later when we examine process memory.

src/core/ruby_version.rs
acj: > I might have missed that - isn't it what we expect THREAD_RUNNABLE to tell? Or does runnable mean "can be run given that it obtains the GVL"? If the latter, then I think that looking at it doesn't make any sense at all.

I think this is a key question, and I don't know enough about RMI VM internals to give a confident answer. What you said seems reasonable. I'll poke around in some reference materials that I've collected and will post again if I find anything.

> Yup. Well, basically, no --on-cpu means --off-cpu now :)

My feeling is that the absence of `--on-cpu` should cause rbspy to collect all traces, which is the current shipping behavior (and what this PR would do in its current state, IIUC), and then we could introduce `--off-cpu` to do the opposite filtering that `--on-cpu` does, i.e. only collect traces if the thread is waiting.

src/core/initialize.rs
michelhe: > I touched on this in another comment, but if this is meant as an initial filter and the more accurate check happens later (via `get_thread_status`), then I think it's okay. Unless the remoteprocess functions we're calling here are expensive, this should reduce rbspy's CPU usage a bit when we're doing on-CPU profiling. It might be worth it to check that assumption, though -- I don't know offhand how expensive this loop is for a typical program.

Anyway, I used the [flame](https://crates.io/crates/flame) crate to insert tracepoints in `get_trace()` and this is what I had measured:
image
```
THREAD: 140546956667136
| get_trace: 0.246437ms
  | is_on_cpu_os_specific: 0.024342ms
  | get_stack_trace_from_current_thread: 0.221337ms
```

It seems like the remoteprocess functions a bit expansive to call per trace..

michelhe: @acj What is your take given this measurement ? should I get rid of the `remoteprocess` calls ?

acj: I don't know yet. On the one hand, the call to check active threads isn't free, but on the other hand we'd avoid the more expensive call to `get_stack_trace_from_current_thread` if the process is definitely not on-CPU. That might be a reasonable tradeoff for some workloads.

Let's revisit this after we solve the correctness/reliability questions.

acj:
I looked through the ruby source and noticed that `rb_thread_struct` has a [`thread_id` field](https://github.com/ruby/ruby/blob/491ab2820aada99f53c328f83f0345acc517fc37/vm_core.h#L963) that we might be able to use to correlate active OS threads with the active thread in the ruby VM. That would involve reading VM memory earlier than we currently do. For example, we could build a list of active OS threads, lock the process, and then immediately get the active thread ID from the VM to see if it's in the active list. If not, then unlock and return. If it's in the active list, then I think we can be reasonably confident that the active thread is on-CPU if its VM thread state is runnable. (This assumes that there's no subtle VM and/or GVL behavior that I'm not accounting for here.)

Thoughts?

michelhe:
Hey @acj ,
I wouldn't want the work done here to become stale,  so please lmk what you think needs to be done in order to merge this upstream 😄

acj:
@michelhe Yeah, we should make a decision here. Are we still blocked on the accuracy/correctness issues? Did you have any thoughts on my last comment?

If you'd rather merge this as an experimental feature and iterate on it later, I'm okay with that too.

Jongy:
> @michelhe Yeah, we should make a decision here. Are we still blocked on the accuracy/correctness issues? Did you have any thoughts on my last comment?

I think that we want to keep the OS-level "is running" check.

About locking/unlocking - I prefer to keep it as accurate as possible also in the nonblocking mode. I recently read Ben's blog at https://www.benfrederickson.com/why-python-needs-paused-during-profiling/ which led me to understand that rbspy doesn't take as long as py-spy to read the stacks. So I suggest that we do something similar to py-spy, that is:
* Lock if blocking
* Check OS level "is running"
* Check Ruby-level "is running" (in py-spy, that's done only if `--gil` is passed - does rbspy have any counterpart or we should always do it? I suppose it should be configurable)
* Read the stacks array
* Unlock if blocking
* Symbolicate the stacks array

acj:
Sorry for the delay. From rereading the above discussion, I think the open questions are:

1. Are we okay with the current heuristic for the OS-level check, i.e. if any thread in the thread group is on-CPU, then the check passes?
2. If that's not accurate enough, then do we want to try what I suggested in my [earlier comment](https://github.com/rbspy/rbspy/pull/327#issuecomment-877696837), i.e. to compare the ruby thread's `thread_id` field with the active OS thread? There may be portability issues here.
3. Depending on how confident we are about 1 and 2, should we treat the `--on-cpu` feature as experimental for now?
4. Does the current code work on FreeBSD? The API seems to have what we need, so I think it should work, but I haven't tried it.

If there's any lingering doubt, I'd vote to deal with 4 (FreeBSD support) and then merge this with `--on-cpu` marked as experimental. Then we can let people try it out, gather more data from our own usage of it, etc.

> Check Ruby-level "is running" (in py-spy, that's done only if --gil is passed - does rbspy have any counterpart or we should always do it? I suppose it should be configurable)

@Jongy We touched on that [here](https://github.com/rbspy/rbspy/pull/327#discussion_r651279237). rbspy doesn't have an equivalent flag right now, but we could add one if we can detect which thread holds the GVL. We might need to ask the ruby VM developers about it.

acj:
@michelhe I'm doing well, thanks for asking. :-) How are you?

> Sorry for taking this long, I'm now ready to begin the needed work to finally bring this feature to mainline.
We actually have been using rbspy from this branch for the past year in our [gProfiler](https://github.com/Granulate/gprofiler) and are generally happy with the results it produces for on-cpu, even though the accuracy with the current state of the code should really be taken with a grain of salt.

That's great to hear. I know that a few rbspy users are interested in this functionality (even if it's experimental) and can help us track down any bugs or quirks. Your proposed course of action sounds good to me.

> Resolve current conflicts with master

Please note that the default branch is now `main`, so you'll need to update any local references.

> Making sure CI passes for FreeBSD - is this enough ?

Yep, that's fine. I can try the feature in my FreeBSD VM after the dust settles. It doesn't need to block merging the PR though as long as CI is happy.

acj:
Ah, sorry for the churn. The parts that we moved into spytools are (currently) for process discovery and memory maps and such, so they shouldn't affect this work much. Let me know if you run into any snags.

acj:
@michelhe :wave: I've freshened this up in a branch [here](https://github.com/rbspy/rbspy/compare/main...acj:rbspy:on-cpu). There are a few minor changes to align with refactors from the past year, clarify that it's experimental, etc. I also reworked the off-cpu test to make sure the timing is reliable. If you're still interested in the feature, please let me know what you think.

michelhe:
> @michelhe :wave: I've freshened this up in a branch [here](https://github.com/rbspy/rbspy/compare/main...acj:rbspy:on-cpu). There are a few minor changes to align with refactors from the past year, clarify that it's experimental, etc. I also reworked the off-cpu test to make sure the timing is reliable. If you're still interested in the feature, please let me know what you think.

Oh wow, thanks for putting in that effort!

IIRC the issues I had with my initial approach was that it was "racy" in some flows - Ruby VM is detected to be on-cpu but by the time the sample is taken, the OS scheduler may have put the process to sleep. We treated it as "good enough" since the small tests we conducted looked good, but this was not validated against a fully fledged application.

Unfortunately, we don't deal with Ruby profiling nowadays 😅 so I cannot say that I'll be able to take it any further or have use for it.
I'd hate to see this code go to waste so as long as it's behind a feature flag and documented to be experimental, I think it'd be a good idea to put it upstream and maybe as it gets used by people it'll improve 🙏

acj:
This was merged as 0f7930f1cc3e0def60428988c4fe75a1b2201db5 and should ship in 0.28.0 later this month. Thanks!

PR #329, opened 2021-07-05
Introduce library crate

### LINKED ISSUE #320: Support using rbspy as a library
There's been some discussion of building rbspy as both a binary and a library, e.g. in #180. This would enable other Rust (and C, maybe others) programs to use rbspy as a native library instead of shelling out to run the binary. The [Pyroscope fork](https://github.com/pyroscope-io/rbspy) of rbspy has already taken a step in this direction.

We also have prior art in py-spy: https://github.com/benfred/py-spy/issues/110.

This shouldn't require too much work. I think we'll want to audit the public surface area of the existing modules to make sure that they're an API we're comfortable maintaining. It's probably okay to make breaking changes until we reach 1.0, but I'd like us to make a reasonable effort to start out with a clean API.

acj:
While rbspy is very useful as a standalone binary, there are use cases that would benefit from using it as a library instead. For example, Pyroscope has been building a statically linked version of rbspy with a simple snapshot interface and calling it from Go code. It's also been suggested that rbspy could be wrapped in a ruby gem to make it easier for ruby programs to profile themselves, e.g. in response to interesting events at runtime.

This PR introduces a small library crate with just enough API surface for the the `main` module to work. Most of the work was moving code from the main module into the library and then cleaning up the module imports. Instead of keeping the long list of recording options as parameters, I introduced a `RecordConfig` struct to hold them. Behind the public API (for now), the core sampling code has been moved to its own module to make the code easier to understand and test.

There are cargo "example" programs for each key feature, which are a good way to see the library API in action.

With a bit more effort, it will be possible to interact with the library from languages like C and Go. (cc @petethepig)

Resolves #320


examples/record.rs
```diff
@@ -0,0 +1,56 @@
...
+    let mut process = std::process::Command::new(path_to_ruby_binary())
+        .arg("ci/ruby-programs/infinite.rb")
+        .spawn()
+        .unwrap();
+    let out_path = std::path::PathBuf::from("rbspy-out.svg");
+
+    let config = RecordConfig {
+        format: OutputFormat::flamegraph,
+        raw_path: std::path::PathBuf::from("rbspy-raw.txt"),
+        out_path: out_path.clone(),
```
petethepig, line 16: One thing I'm thinking about is it would be nice to have some sort of a way to exchange these directly in memory and thus avoid overhead when a) serializing profiling data and b) writing it to file.

I don't know how to do that though, especially in a way that would work with a simpler C api.

@AdrK, what do you think? Would it have to be a char* buffer that users provide or something like that?

src/sampler/record.rs
```diff
@@ -0,0 +1,451 @@
...
+const BILLION: u64 = 1000 * 1000 * 1000; // for nanosleep
+
+// This SampleTime struct helps us sample on a regular schedule ("exactly" 100 times per second, if
+// the sample rate is 100).
+// What we do is -- when doing the 1234th sample, we calculate the exact time the 1234th sample
+// should happen at, which is (start time + nanos_between_samples * 1234) and then sleep until that
+// time
+struct SampleTime {
+    start_time: Instant,
+    nanos_between_samples: u64,
```
AdrK, line 26: NIT
Why unsigned? Signed variables are generally better handled by the compiler as the compiler can do more assumptions thus allowing for more optimization. I would prefer unsigned only in places where we deal with binary content rather than the value itself.
I guess that the reasoning here is that there is a need to have huge range as we are dealing with nanoseconds, but int64 is easily able to hold values of many years in nanoseconds.
http://blog.llvm.org/2011/05/what-every-c-programmer-should-know.html#signed_overflow

src/sampler/record.rs
```diff
@@ -0,0 +1,451 @@
...
+                sender.send(ok_trace)?;
+            }
+            Err(x) => {
+                if let Some(MemoryCopyError::ProcessEnded) = x.downcast_ref() {
+                    debug!("Process {} ended", pid);
+                    return Ok(());
+                }
+
+                errors += 1;
+                if errors > 20 && (errors as f64) / (total as f64) > 0.5 {
```
AdrK, line 379: What if there will be less than 20 errors? I guess those would not be printed. Maybe there should be a check if any error occurred, just after the loop and then they can be printed?

examples/record.rs
acj: I'll keep this in mind. It seems doable for individual snapshot data, but I don't see an obvious path for recording.

src/sampler/record.rs
acj: This is old code that got moved, so I'm not sure. It might have been an arbitrary choice.

acj:
@michelhe Oh, interesting. What use case do you have in mind?

src/sampler/record.rs
acj: Yeah, we don't have a nice way to report non-fatal errors right now. Would be nice. I'll keep this in mind too.

examples/record.rs
AdrK: Imo, the best would be to use shared memory approach, Linux has a shm lib for this purpose. Shared file is quite slow apporach, as the open, seek and close operations are expensive. If we can assume that writer & reader will always be in the same process, we can simply pass a pointer to the memory between those.

michelhe:
> @michelhe Oh, interesting. What use case do you have in mind?

For example, also exposing the remoteprocess through a convinent API and using the structs crate to hack around and make manipulations to a live ruby VM.

I'm not sure if rbspy by itself is the right platform for such a thing but it already has all the building blocks already implemeted ^^

acj:
@michelhe Gotcha. That's a neat idea, though I think it should be separate from rbspy. The ruby-structs crate only has the types that rbspy needs (see [`scripts/bindgen.sh`](https://github.com/rbspy/rbspy/blob/521878a5b289f35fc9b04e0cc660464807276a67/scripts/bindgen.sh)) and would be fairly limited, but it should be straightforward to copy/adapt ruby-structs and other parts of rbspy as needed. If you do build this, I hope you share it -- it sounds fun and useful. :)

acj:
Update: I haven't merged this yet because I wanted to explore a set of changes/refactors that would separate the CLI, recording, and sampling code. In addition to separating distinct concerns, I'm hoping it will make the library API cleaner. It's going well and should be in a sharable state soon, maybe this week or next week.

examples/record.rs
acj: Now that the components are separated, we could consider exporting the `sampler` as a way to get low-level access to traces. A program like Pyroscope could include a small shim (probably in Rust) that creates a sampler and starts it up, collecting traces into batches and passing them in memory to C/Go/whatever code that consumes them. I'd prefer not have that complexity in rbspy itself, but it seems reasonable for rbspy's library crate to provide the tracing infrastructure to support it.

Issue #334, opened 2021-07-22
bug: Early exits, duration parameter not respected in some cases


dalehamel:
In some cases, we see that rbspy exits early, both with and without the `--duration` parameter.

Without the duration parameter, it can sometimes exit with an exit code of 0. The target process has not exited, and there is no apparent error in the output.

With the duration parameter, we can sometimes see it exit well before the duration has elapsed:

```
$ ./rbspy-x86_64-unknown-linux-gnu record -p 908389 -d 120

Time since start: 31s. Press Ctrl+C to stop.
Summary of profiling data so far:
% self  % total  name
 81.04    81.14  select [c function] - (unknown)
  9.38     9.38  pop [c function] - (unknown)
  2.42     2.42  sleep [c function] - (unknown)
  0.77     0.77  wait_readable [c function] - (unknown)
  0.77     0.77  initialize [c function] - (unknown)
  0.58     0.58  _query [c function] - (unknown)
  0.45     2.00  new [c function] - (unknown)
  0.19     0.52  initialize - /artifacts/bundle/ruby/2.7.0/gems/graphql-1.12.13/lib/grap
  0.19     0.19  tcp_listener_stats [c function] - (unknown)
  0.13     0.13  split [c function] - (unknown)
  0.10     0.10  match? [c function] - (unknown)
  0.06     5.58  public_send [c function] - (unknown)
  0.06     3.93  block in run - /app/app/models/engine.rb
  0.06     1.39  block in coerce_into_values - /artifacts/bundle/ruby/2.7.0/gems/graphql
  0.06     0.68  synchronize [c function] - (unknown)
  0.06     0.48  block in rules_from_db - /app/app/models/ordered_ruleset.rb
  0.06     0.06  legacy_connection_handling - /artifacts/bundle/ruby/2.7.0/gems/activere
  0.06     0.06  inheritance_column - /artifacts/bundle/ruby/2.7.0/gems/activerecord-6.1
  0.06     0.06  eval [c function] - (unknown)
  0.06     0.06  clock_gettime [c function] - (unknown)
1.8% (56/3109) of stack traces were sampled late because we couldn't sample at expected rate, results may be inaccurate. Current rate: 100. Try sampling at a lower rate with `--rate`.
Wrote raw data to /root/.cache/rbspy/records/rbspy-2021-07-22-1cem3j2wTP.raw.gz
[...]

PR #337, opened 2021-07-26
Allow retrying process locking in case of spurious issues

dalehamel:
Attempt to fix https://github.com/rbspy/rbspy/issues/334

# Problem being solved

It appears that in some cases attempting to lock the process will fail, when running rbspy without `--nonblocking`. This is fairly easy to reproduce by setting the sample rate to something very high (500 or 1000), as this increases the likelyhood that it will fail to lock one of the threads associated with the process.

It looks like the [implementation of `lock`](https://docs.rs/remoteprocess/0.4.3/src/remoteprocess/linux/mod.rs.html#65-86) is inherently racy on linux because it interacts with the system when trying to lock a thread via the call to ptrace, it is possible that the **thread** will no longer exist, and ptrace will correctly return ESRCH in this case. However, we treat failing to lock the process due to this error as if the process itself does not exist, which leads to the whole tracer exiting, thinking the target process is no longer running when it in fact still is.

# Solution

To resolve this problem, as @acj has suggested in https://github.com/rbspy/rbspy/issues/334#issuecomment-885937078, I introduce some retry logic:

- The current locking code is moved to a closure, `try_lock`, so that it can be easily called repeatedly
- A new type of MemoryCopyError specific to this issue is introduce, `ProcessNotLocked`, which is returned by the closure if we detect that we got ESRCH, but the process still exists
- We will retry locking up to 10 times, but only if we received `ProcessNotLocked` each time - on any other type of error, the retry loop will be broken immediately. In practice one retry is probably sufficient, and it is highly unlikely that it should hit his error ten times in a row, but ten seems like a reasonable upper limit to the retries.
- Even if all retries were exhausted, the error is still distinguished from `ProcessEnded`, so the original bug of this tripping https://github.com/rbspy/rbspy/blob/master/src/main.rs#L702 should be fix; this error will be treated as any other error, and not claim a successful exit

Using the modifications here, I was able to profile a production puma process for 120 seconds uninterrupted, at 500hz, which I cannot do without this patch.

Note - i am very unfamiliar with Rust, this is just my first attempt at a solution as I am also trying to maneuver my way around the language. All feedback including nitpicks and style suggestions, and comments on standard idiom are sincerely welcome.

Since the check suite doesn't run for new contributors, here is the run on my own fork: https://github.com/dalehamel/rbspy/runs/3168336700?check_suite_focus=true

### DIFF
--- a/src/core/initialize.rs
+++ b/src/core/initialize.rs
@@ -104,7 +104,7 @@ impl StackTraceGetter {
                              Ok(remoteprocess::Error::NixError(e)) => match e {
                                  nix::Error::Sys(nix::errno::Errno::EPERM) => {
                                      debug!("EPERM for process {} : {:?}", self.process.pid, e);
-                                     return Err(MemoryCopyError::ProcessNotLocked);
+                                     return Err(MemoryCopyError::PermissionError);
                                  }
                                  nix::Error::Sys(nix::errno::Errno::ESRCH) => {
                                      match Process::new(self.process.pid) {
@@ -142,10 +142,17 @@ impl StackTraceGetter {
                         break;
                     }
                     Err(e) => match e {
-                        MemoryCopyError::ProcessNotLocked => {
+                        MemoryCopyError::ProcessNotLocked | MemoryCopyError::PermissionError => {
                             if retries == 0 {
                                 debug!("Exhausted retries to lock process");
-                                return Err(e)?;
+                                match e {
+                                    MemoryCopyError::PermissionError => {
+                                        return Err(MemoryCopyError::ProcessEnded)
+                                    }
+                                    _ => {
+                                        return Err(e)?;
+                                    }
+                                }
                             }
                             debug!("Retrying lock");
                             retries -= 1;
--- a/src/core/initialize.rs
+++ b/src/core/initialize.rs
@@ -134,7 +134,10 @@ impl StackTraceGetter {
             loop {
                 match try_lock(&self.process) {
                     Ok(l) => {
-                        debug!("Was able to lock process {} ", self.process.pid);
+                        debug!(
+                            "Was able to lock process {}, {} retries remained ",
+                            self.process.pid, retries
+                        );
                         _lock = l;
                         break;
                     }
@@ -425,6 +428,42 @@ fn test_get_exec_trace() {
         getter.reinit_count, 1,
         "Trace getter should have detected one reinit"
     );
+}
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_thread_churn() {
+    // Test that even a program churning threads can be locked to be traced
+    // Success / error criteria here should match those from `fn record` main.rs
+    let mut process = std::process::Command::new("ruby")
+        .arg("./ci/ruby-programs/thread_churn.rb")
+        .spawn()
+        .unwrap();
+    let pid = process.id() as Pid;
+    let mut getter = initialize(pid, true).unwrap();
+    std::thread::sleep(std::time::Duration::from_millis(50));
+
+    let total_samples = 1000;
+    let mut errors = 0;
+    for _ in 0..total_samples {
+        std::thread::sleep(std::time::Duration::from_millis(1));
+        match getter.get_trace() {
+            Ok(_t) => {}
+            Err(e) => match e.downcast_ref() {
+                Some(MemoryCopyError::ProcessEnded) => {
+                    panic!("Spurious process exit while testing thread churn");
+                }
+                _ => {
+                    errors += 1;
+                }
+            },
+        }
+    }
+
+    // These assertions are based on those from the record function in main.rs
+    assert!(errors <= 20);
+    assert!((errors as f64) / (total_samples as f64) <= 0.5);
+    process.kill().unwrap();
 }
 
 #[test]
--- a/src/core/initialize.rs
+++ b/src/core/initialize.rs
@@ -80,40 +80,73 @@ impl StackTraceGetter {
     fn get_trace_from_current_thread(&self) -> Result<StackTrace, MemoryCopyError> {
         let stack_trace_function = &self.stack_trace_function;
 
+        let try_lock = |process: &Process| {
+            process
+                 .lock()
+                 .context("locking process during stack trace retrieval")
+                 .or_else(|e| {
+                     if let Some(source) = e.source() {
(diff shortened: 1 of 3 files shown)


dalehamel:
Sorry for all the pings, PTAL when you have a chance @acj

src/core/initialize.rs
```diff
@@ -80,40 +80,73 @@ impl StackTraceGetter {
...
-                                    std::io::ErrorKind::NotFound => {
-                                        return Err(MemoryCopyError::ProcessEnded)
-                                    }
-                                    // Windows
-                                    std::io::ErrorKind::PermissionDenied => {
-                                        return Err(MemoryCopyError::ProcessEnded)
-                                    }
-                                    _ => {}
-                                };
+            let mut retries = 10;
```
dalehamel, line 130: retries should be very uncommon, as they would necessitate the lock failing on ESRCH repeatedly, so each successive occurrence of this should be less and less probable. In reality, one retry is probably enough most of the time, and only in exceptional circumstances should the 10 retries here be hit

acj:
This seems reasonable to me. Thanks for the patch. I'll take a closer look in the next few days and will put it through its paces.

> This is fairly easy to reproduce by setting the sample rate to something very high (500 or 1000), as this increases the likelyhood that it will fail to lock one of the threads associated with the process.

Have you found a more minimal repro case than profiling a busy production app? I haven't seen the issue in my local puma setup while generating traffic with `ab`, but that might be too uniform and/or not causing enough thread churn. I might try writing a test script that starts and stops a bunch of threads. Would be great to capture this in a test somehow.

dalehamel:
Thanks for taking a look @acj 

> Have you found a more minimal repro case than profiling a busy production app?

Unfortunately no.

We noticed this issue while rolling out rbspy from our lower-tier application clusters (which are less busy) to our more busy clusters, higher-tier clusters. The lower tier clusters would run for their full duration without any issues. Once we put rbspy in our higher-tier clusters, we noticed that some traces would exit after only 4 or 10 seconds, rather than the 30-120 seconds requested. As I mentioned in the issue, `--nonblocking` prevents the problem entirely, so this plus the `ESRCH` error is what lead me to this solution.

So it definitely has something to do with traffic, and I can say with certainty that the case I was using repeatedly was a puma worker in cluster mode, but beyond that I don't have a more minimal example, unfortunately.

I would have liked to try and write a test to detect this issue, but since it comes down to (I think) a timing problem, it comes down to probability, and I think that higher sampling rate + higher traffic was what lead to us seeing it frequently.

dalehamel:
Converted back to draft because @acj has come up with a good test program which this doesn't pass on https://github.com/rbspy/rbspy/issues/334#issuecomment-889575929

src/core/initialize.rs
dalehamel: From the test case that @acj wrote, I found I needed to bump this to 20 retries to pass the test. I don't expect it is likely to see such pathological cases in the wild, but with 20 we should be confident that the retries will be sufficient.

src/core/initialize.rs
```diff
@@ -80,40 +80,83 @@ impl StackTraceGetter {
...
+                                         return Err(MemoryCopyError::ProcessEnded)
+                                     }
+                                     _ => {}
+                                 };
+                             }
+                             #[cfg(target_os = "linux")]
+                             Ok(remoteprocess::Error::NixError(e)) => match e {
+                                 nix::Error::Sys(nix::errno::Errno::EPERM) => {
+                                     debug!("EPERM for process {} : {:?}", self.process.pid, e);
+                                     return Err(MemoryCopyError::ProcessNotLocked);
```
dalehamel, line 107: I've changed this from ProcessEnded, as I think this is a more appropriate error type - we couldn't lock the process, and got EPERM back. Since this is retriable, we may as well retry, which is the handling for ProcessNotLocked.

src/core/initialize.rs
```diff
@@ -80,40 +80,90 @@ impl StackTraceGetter {
...
+                        _lock = l;
+                        break;
+                    }
+                    Err(e) => match e {
+                        MemoryCopyError::ProcessNotLocked | MemoryCopyError::PermissionError => {
+                            if retries == 0 {
+                                debug!("Exhausted retries to lock process");
+                                match e {
+                                    MemoryCopyError::PermissionError => {
                                         return Err(MemoryCopyError::ProcessEnded)
```
dalehamel, line 150: This is a bit awkward, but is to mimic existing behavior as to not break the `test_spawn_record_children_subprocesses` test, but I believe that this is actually a bug in the other teast, which was previously swallowing permission errors, treating them as though the program had exited successfully.

Were it not for this, we could treat EPERM as a ProcessNotLocked error, but we need to distinguish between the two to preserve compatibility.

We should probably file a separate bug for this EPERM leading to a 0 exit code that has also been uncovered by this test case.

src/core/types.rs
```diff
@@ -42,6 +42,10 @@ pub enum MemoryCopyError {
     Io(usize, std::io::Error),
     #[error("Process isn't running")]
     ProcessEnded,
+    #[error("Permission error")]
+    PermissionError,
```
dalehamel, line 46: See above comment for description of why this new error type is added

dalehamel:
@acj I added your test program from #337 (noted you as co-author, but in the commit gave you full credit - hope that's OK), and got it to succeed on my own actions run: https://github.com/dalehamel/rbspy/runs/3199001859?check_suite_focus=true

A few things to note:

- The test case also made another error apparent → EPERM also leads to a spurious exit. As I node above, I think that existing tests are swallowing this error. I had to modify the logic to preserve the existing behavior to not break tests.
- I had to up the max retries from 10 to 20 for the really pathological test case of your thread churning + 1000 samples at 1000hz to pass
- The test case I made is based on the error criteria in `record` from main.rs, since it is possible to get other memory copy errors while reading a stack trace. As long as it is within the acceptable threshold for errors, it is deemed to have passed, unless a ProcessEnded error is surfaced.

src/core/initialize.rs
dalehamel: I ended up having to make a new error type for this, since I think another test depends on treating EPERM as ProcessEnded, which is ultimately treated as Ok downstream, though I think this leads to swallowing an error incorrectly.

dalehamel:
Marking as a draft again because the test seems flakey

acj:
The most recent failure looks like a quirk in the FreeBSD VM. I kicked off a fresh build.

Issue #333, opened 2021-07-18
[linux] Buggy tests, processes not killed


Horki:
Kernel Release: 5.12.15-arch1-1

On Linux, sometimes when I run tests, and some test fails, due to panics, `process::kill` is never invoked, and they keep running in the background.

PS This also happens if I enforce to run single-threaded tests via`cargo test --release -- --test-threads=1`

`% cargo test --release`

```
running 27 tests
test core::initialize::test_initialize_with_disallowed_process ... ok
test core::initialize::test_initialize_with_nonexistent_process ... ok
test core::ruby_version::tests::test_get_ruby_stack_trace_1_9_3 ... ok
test core::ruby_version::tests::test_get_ruby_stack_trace_2_4_0 ... ok
test core::ruby_version::tests::test_get_ruby_stack_trace_2_5_0 ... ok
test core::ruby_version::tests::test_get_ruby_stack_trace_2_1_6 ... ok
test core::initialize::test_get_trace ... FAILED
test core::initialize::test_get_exec_trace ... FAILED
test core::initialize::test_current_thread_address ... FAILED
test test_output_filename ... ok
./ci/ruby-programs/ruby_exec.rb:4:in `readline': end of file reached (EOFError)
        from ./ci/ruby-programs/ruby_exec.rb:4:in `<main>'
test core::ruby_version::tests::test_get_ruby_stack_trace_2_7_2 ... ok
test tests::test_arg_parsing ... ok
test tests::test_report_arg_parsing ... ok
test ui::callgrind::tests::stats_aggregate ... ok
test ui::callgrind::tests::stats_write ... ok
test ui::flamegraph::tests::test_collapsed ... ok
test ui::flamegraph::tests::test_flamegraph_from_collapsed ... ok
test ui::flamegraph::tests::test_stats ... ok
test ui::summary::tests::stats_by_function ... ok
test ui::summary::tests::stats_by_line_number ... ok
test core::ruby_version::tests::test_get_ruby_stack_trace_2_7_4 ... ok
test core::ruby_version::tests::test_get_ruby_stack_trace_3_0_0 ... ok
test core::ruby_version::tests::test_get_ruby_stack_trace_2_7_3 ... ok
test core::ruby_version::tests::test_get_ruby_stack_trace_3_0_2 ... ok
test core::ruby_version::tests::test_get_ruby_stack_trace_3_0_1 ... ok
test core::ruby_version::tests::test_get_ruby_stack_trace_2_1_6_2 ... ok
[...]

PR #338, opened 2021-08-04
Handle permission errors while trying to lock

dalehamel:
If a permission error is encountered while trying to lock the target process,
the, returning ProcessEnded is a misrepresentation of the error: the process
has not ended, we failed to lock it due to a permission error.

This matters because ProcessEnded is the error condition that leads to a
successful exit, so permission errors may result in a premature exit.

This commit also fixes a logical error in test_spawn_record_children_subprocesses
which could result in the test itself causing an EPERM issue to arise on Linux,
as it would cause the root process that it spawns to be defunct, leading to
EPERM while trying to use ptrace to lock it. Moving the process wait earlier in the test avoids this issue.

This is described further in https://github.com/rbspy/rbspy/issues/334#issuecomment-892977927 and https://github.com/rbspy/rbspy/issues/334#issuecomment-892991127


src/core/initialize.rs
```diff
@@ -96,15 +96,15 @@ impl StackTraceGetter {
                                     }
                                     // Windows
                                     std::io::ErrorKind::PermissionDenied => {
-                                        return Err(MemoryCopyError::ProcessEnded)
+                                        return Err(MemoryCopyError::ProcessNotLocked)
```
dalehamel, line 99: I'm speculating that this issue may also be present on windows, either way I think ProcessEnded is not a good error to provide in this case, and ProcessNotLocked is more descriptive.

dalehamel:
Passes locally in one-off tests, but errors in full test suite and in CI so marking as draft

dalehamel:
It looks like the tests are order-dependent, and I'm noticing other defunct processes, probably related to https://github.com/rbspy/rbspy/issues/333

dalehamel:
I got this to pass locally so I don't think the suite failures are due to this PR, marking back as ready-to-review, but it may still take a few test runs in CI since they seem to be order-dependent.

dalehamel:
test keeps timing out in CI, I'll try debugging why this is.

dalehamel:
I discovered another issue with defunct processes - when we are given a command to run, such as in our malloc tests, we never wait for the subprocess, and it stays a zombie / defunct. This leads to the EPERM issue again. Working on a fix for this as well.

src/core/initialize.rs
acj: I'll test this before merging. I recall that process exit on Windows can result in an EPERM-like error code, but it might happen when we're walking through VM memory and not here.

src/main.rs
```diff
@@ -562,6 +562,15 @@ fn test_spawn_record_children_subprocesses() {
         }
     }
 
+    // It is possible that the spawned children can become defunct, which may
```
acj, line 565: This is puzzling because all of the children should be waited for in ruby_forks.rb. How frequently have you seen this after you moved the `wait` call above the results loop?

acj:
> I discovered another issue with defunct processes - when we are given a command to run, such as in our malloc tests, we never wait for the subprocess, and it stays a zombie / defunct. This leads to the EPERM issue again. Working on a fix for this as well.

In case it's helpful, here's a WIP commit based on work I'm doing for another issue: https://github.com/acj/rbspy/commit/aa8d0dac3908372cab965ca09e199a9e1c20cd3d. Feel free to steal/adapt.

dalehamel:
@acj that patch is a HUGE help, i was struggling to try and do the same thing. I'll definitely ~~steal~~ borrow this!

dalehamel:
I think that somewhere in the path of `spawn_recorder_children` we have dangling threads and it is becoming an issue in some edge case. I wonder if we should return a join handler, to join on the subprocess that is recording the children?

I have a patch that seems to prevent the defunct process in **most** cases by waiting on the child process in a background thread, but in some rare cases it blocks indefinitely, so this is what I am chasing down now.

acj:
Thanks for your effort on this. I don't have any insights to offer at this point, but I wanted to mention that the subprocess test has been flaky for a long time, and we'll probably need to chase down quite a few edge cases to make it reliable. It's fine if you want to land a narrow version of these fixes and then iterate on the rest separately. Or, if you're enjoying the hunt, we can keep trying to find them.

> I wonder if we should return a join handler, to join on the subprocess that is recording the children?
>
> I have a patch that seems to prevent the defunct process in most cases by waiting on the child process in a background thread, but in some rare cases it blocks indefinitely, so this is what I am chasing down now.

Yeah, this makes me wonder if we're not detecting a child exit, e.g. due to another race condition. When I've worked on the subprocess test in the past, deadlocks have usually meant that we ended up in a situation where one or more child threads didn't close their channels because they still thought their process was alive (or they mishandled an error), and so the top-level recorder thread kept running. This feels similar.

Which platform are you doing your testing on?

dalehamel:
@acj sorry for the delay, was under the weather last week. Hope to return to this soon.

> Which platform are you doing your testing on?

Linux x86_64 primarily (local laptop, running in a VM), and then github actions for the matrix tests.

>  It's fine if you want to land a narrow version of these fixes and then iterate on the rest separately.

So the kicker is that this PR has uncovered a regression - it shows a couple of tests that were passing that probably should have been failing, almost all related to defunct processes in some way or another. I'm not sure how to land this PR without "fixing everything" as it were.

> Or, if you're enjoying the hunt, we can keep trying to find them.

It's proving to be an interesting exercise, and a great introduction to Rust for me. If there isn't a rush on your end, I'd like to try and unravel it completely and come up with a solution that both fixes the behavior, and addresses the regressions.

acj:
> sorry for the delay, was under the weather last week. Hope to return to this soon.

No problem at all. I hope you're feeling better!

> It's proving to be an interesting exercise, and a great introduction to Rust for me. If there isn't a rush on your end, I'd like to try and unravel it completely and come up with a solution that both fixes the behavior, and addresses the regressions.

Great, I'm glad to hear that. There's no rush. As a heads up, I'm hoping to merge #329 soon (next few days), which will shuffle things around a bit but should have a minimal impact on behavior. My hope is that the refactored module layout will also make these process/thread bugs easier to trace and debug.

acj:
I was running into similar test failures in #329 and ended up cherry-picking part of the first commit from this PR (with credit to you) to fix them, or at least to make the tests more stable. The other refactoring work shouldn't affect behavior or timing very much, though the sampling setup process is slightly different. The rest of your changes should rebase cleanly, I think.

acj:
:wave: I'm cleaning up older PRs and issues. If you're still interested in this, feel free to reopen.

PR #344, opened 2021-11-23
Add pprof output format

saunderst:
This adds a new outputter to support the trace output format described here: https://github.com/google/pprof/tree/master/proto.  This is a format typically used to profile golang programs.  It can be used to upload traces to [Google Cloud Profiler](https://cloud.google.com/profiler), and [Speedscope also supports it](https://github.com/jlfwong/speedscope/wiki/Importing-from-pprof-(go)).

I've closely followed the approach used to add Speedscope support, so most of it should be fairly self-explanatory.

~In the first commit, I've also done some refactoring in `Recorder::record()` to expand the API a bit without changing functionality.  This was helpful in implementing an additional `Recorder` trait for another project I'm working on.  If this change is undesirable, the `pprof`-related code should work the same without it.~ (See comment below)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -158,6 +158,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "14c189c53d098945499cdfa7ecc63567cf3886b3332b312a5b4585d8d3a6a610"
 
 [[package]]
+name = "bytes"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c4872d67bab6358e59559027aa3b9157c53d9358c51423c17554809a8858e0f8"
+
+[[package]]
 name = "cc"
 version = "1.0.72"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


saunderst:
Just pushed a little tweak to get the test running in Windows.

acj:
Thanks for working on this! The pprof changes look good aside from the test issues.

I'd prefer to remove the first commit since it's not related to pprof support. I'm happy to consider the API changes as a separate thing, though, if there's a clear need. It would be helpful to hear more about your other project and how it's using rbspy, if you're able to share that info.

saunderst:
Oops -- test should be good now!

The other project is doing continuous recording with periodic output to various configurable destinations.  I needed access to some of the `Recorder` fields, so rather than make them all public (which would have been easier, but maybe not the "right" way to do it?), I exposed the two new methods I needed, and then (just for consistency and as a first check on my work) I refactored `Recorder::record()` to use them as I'm using them in my project.  Anyway, I'm not married to this approach.  I'd be interested in agreeing on some way to do this so that I could completely eliminate my `rbspy` fork.  If not by exposing `Recorder` fields, maybe by upstreaming my continuous recording method?  It's a bit tailored to my needs right now, but maybe it could be generalized if it would be useful for `rbspy`.  Either way, let me know what you think.  Thanks!

acj:
Ah, that's an interesting use case. The public API is currently very small -- just enough for the rbspy CLI to function -- but it could be made a bit larger to accommodate projects like yours. For example, if we make the `sampler` module part of the public API, and probably also the `ui` module, then you could make your own recorder that directly interacts with those components. How does that sound?

After this next release (maybe later today), rbspy will be available as a library crate as well as a binary one, which should make it easier for you to package your project without maintaining an rbspy fork.

Anyway, I'm happy to continue discussing this. If my suggestion above sounds good to you, please open an issue for the API changes so that we can keep the conversation in one place.

acj (approved):
Thanks!

PR #348, opened 2021-12-20
Enhance the user experience in pprof-consuming apps by adjusting units

saunderst:
Although any arbitrary string can be used for `sample_type` units in pprof format, I've accidentally learned that the user experience in Google Cloud Profiler and Speedscope can be improved by specifying these units as "nanoseconds".  When using "ns" we see large numbers with a lot of commas everywhere, but by changing to "nanoseconds" we get automatically scaled display units like "ms" and "s".

Before and after (different samples):

Screen Shot 2021-12-19 at 4 21 09 PMScreen Shot 2021-12-19 at 4 20 25 PM

### DIFF
--- a/src/ui/pprof.rs
+++ b/src/ui/pprof.rs
@@ -25,8 +25,12 @@ impl Stats {
     pub fn new() -> Stats {
         Stats {
             profile: Profile {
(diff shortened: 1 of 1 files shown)


acj (approved):
Thanks!

PR #352, opened 2022-01-18
Update README.md

jacoblockard99:
It looks like this link is broken. Should it be changed to this?


acj (approved):
Yes, thanks! I just overhauled the docs and forgot to update this.

PR #353, opened 2022-02-17
refactor: export StackTrace type

omarabid:
Any reason why these types are private? I'm integrating rbspy with https://github.com/pyroscope-io/pyroscope-rs and they could be useful. So far, I'm using StackTrace, but I think more types could be exposed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,6 +41,7 @@ pub mod ui;
 
 pub use crate::core::process::Pid;
 pub use crate::core::types::OutputFormat;
+pub use crate::core::types::StackTrace;
 
 /// Generate visualization (e.g. a flamegraph) from raw data that was previously recorded by rbspy
 pub fn report(


acj:
Hey, thanks for the PR.

> Any reason why these types are private?

We took a private-by-default approach for the library crate's API, and so the public parts are mostly things that the rbspy binary needs. It's fine to expose more types when it would make the library more useful, though. Do you think you'll need StackFrame too?

omarabid:
@acj Thanks for the clarification. I don't think I'll be using the StackFrame in my current implementation. That being said, I think both should be public as they go (let me know if I should edit my PR)

The reason for that is quite simple. If you are using Sampler::start(...), you need to pass a Channel Sender and get (control) a Receiver. You might get away from defining the type for your Receiver as Rust will infer the type automatically but if you are passing the Receiver around, Rust is not that smart. At that point, you need to declare the proper type (which you can't do!).

For StackFrame, I think it makes sense as it is part of StackTrace. Someone might struggle to define a Vector of StackFrame"s" as he'll need Rust to infer the type for all the variables that use StackFrame.

Hope that makes sense :)

acj (approved):

PR #357, opened 2022-04-03
Add --force-version flag for testing with pre-release rubies

acj:
Testing new versions of ruby before they're released is currently a bit of a pain. You need to mention the new version string(s) in a few places, run bindgen, and recompile. However, breaking changes to the ruby C types that rbspy cares about are fairly rare, and using the types from the previous ruby version usually works fine. This PR adds a `--force-version` flag that overrides the version detection that rbspy normally performs.

The flag might also be useful for anyone maintaining a custom ruby build that has a unique version string but is otherwise compatible with upstream ruby.

For example, I have ruby 3.1.1 installed on my computer, and today I wanted to check rbspy's compatibility with [3.2.0 preview 1](https://www.ruby-lang.org/en/news/2022/04/03/ruby-3-2-0-preview1-released/) that was just released. With the new flag, it's easy:

```
$ ruby -v
ruby 3.2.0dev (2022-04-03T15:09:54Z master fb5aa31e2d) [arm64-darwin21]

$ ruby ci/ruby-programs/infinite.rb &
[1] 84631

$ sudo ./target/debug/rbspy snapshot --pid $(pgrep ruby)
Password:
thread 'main' panicked at 'Ruby version not supported yet: 3.2.0. Please create a GitHub issue and we'll fix it!', src/core/initialize.rs:451:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

$ sudo ./target/debug/rbspy snapshot --pid $(pgrep ruby) --force-version 3.1.1
<main> - /Users/acj/proj/rbspy/rbspy/ci/ruby-programs/infinite.rb:13
loop [c function] - (unknown):0
block in <main> - /Users/acj/proj/rbspy/rbspy/ci/ruby-programs/infinite.rb:15
ccc - /Users/acj/proj/rbspy/rbspy/ci/ruby-programs/infinite.rb:11
bbb - /Users/acj/proj/rbspy/rbspy/ci/ruby-programs/infinite.rb:7
aaa - /Users/acj/proj/rbspy/rbspy/ci/ruby-programs/infinite.rb:3
sleep [c function] - (unknown):0
```

As a followup step, I'd like to set up a CI job that runs ~weekly against ruby's development branch so that we'll be warned about breaking changes well before a release.

### DIFF
--- a/examples/record.rs
+++ b/examples/record.rs
@@ -20,6 +20,7 @@ fn main() {
         maybe_duration: Some(std::time::Duration::from_secs(1)),
         flame_min_width: 10.0,
         lock_process: true,
+        force_version: None,
     };
     let recorder = Recorder::new(config);
     match recorder.record() {
--- a/examples/snapshot.rs
+++ b/examples/snapshot.rs
@@ -9,7 +9,7 @@ fn main() {
         .unwrap();
     let pid = process.id() as rbspy::Pid;
 
-    match snapshot(pid, true) {
+    match snapshot(pid, true, None) {
         Ok(s) => println!("{}", s),
         Err(e) => println!("Failed to get snapshot: {:?}", e),
     }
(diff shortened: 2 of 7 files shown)


saunderst:
It could be worth advertising this feature in the "Ruby version not supported yet" messages.

Issue #372, opened 2022-12-13
Add native traces to `RubySpy::get_stack_trace`


Maaarcocr:
I've been needing to profile a ruby application that uses a variety of ruby extensions (written in C & rust) and also knowing how time is spent in the native extension a lot of the logic resides there.

I've taken inspiration from https://www.benfrederickson.com/profiling-native-python-extensions-with-py-spy/ and their source code to figure out how I could do it with rbspy. I have a prototype that works, and I have 2 questions: 

When I get a StackTrace from ruby I also get a thread id, but it does not seem to match any thread id from `remoteprocess::Process::threads` in `RubySpy`. Is that expected? Is there a way to get a current thread ID that matches one from `remoteprocess`? 

Would there be any interest in merging this into main here? Right now my changes live in my fork. I've made it such this capability is only turned on if a feature flag is set to true.

acj:
Hey, thanks for working on this.

> I have a prototype that works

🎉 

> When I get a StackTrace from ruby I also get a thread id, but it does not seem to match any thread id from remoteprocess::Process::threads in RubySpy. Is that expected? Is there a way to get a current thread ID that matches one from remoteprocess?

Hmm, I'm not sure. The thread ID in the StackTrace is read from `rb_thread_struct`, which might be an internal ID that the Ruby VM controls rather than the OS's thread ID that remoteprocess returns. Can you say more about the problem you're trying to solve? Also, which ruby version are you using to test this?

There's some structural similarity between this and the work in #327 -- might be worth a look.

> Would there be any interest in merging this into main here?

Yes, it would be great to support this capability. Feel free to open a draft PR when you're ready, and then we can work out the details. Worst case, we can keep it behind the feature flag until it's working reliably.

Maaarcocr:
> Can you say more about the problem you're trying to solve? Also, which ruby version are you using to test this?

right so, currently (this is my understanding of rbspy, it may be wrong) rbspy would get the traces only for the current thread every time it samples the ruby process. Ideally I would like to do the same with native traces, i.e. only get the native traces for the thread which is marked as active by the RubyVM and just ignore all the other system threads. 

If I do not do this, then it becomes difficult to know how to interweave the native trace and the ruby trace, as we could be in a situation where we have multiple native traces (one for each OS thread from remoteprocess) and just 1 ruby trace. 

to make my prototype work I currently just assume that there is only 1 thread, but that would not work in the general case.

Maaarcocr:
[...]

PR #373, opened 2022-12-15
Implement native profiling

Maaarcocr:
resolves https://github.com/rbspy/rbspy/issues/372

# How does it work? 

When the `native-profiling` flag is on and we are on linux or windows (remoteprocess only supports unwinding on these 2 platforms afaik) then, after getting the ruby trace, we: 

- get native stack frames using remoteprocess
- go through the ruby frames, when a c function is found
  - find the closest start of a c function invocation in the native frames
  - add all the frames from it
  - stop adding when we know we are going back to ruby

The start of a c function is: `rb_vm_exec` and the end of it is: `vm_call_cfunc_with_frame`.

By interweaving native and ruby frames this way, we get to profile both ruby and native code!

### DIFF
--- a/src/core/ruby_spy.rs
+++ b/src/core/ruby_spy.rs
@@ -3,10 +3,12 @@ use anyhow::format_err;
 use anyhow::{Context, Error, Result};
 use spytools::ProcessInfo;
 
-use crate::StackFrame;
-use std::time::{Duration, Instant};
 use crate::core::process::{Pid, Process, ProcessRetry};
 use crate::core::types::{MemoryCopyError, StackTrace};
+use crate::StackFrame;
+use std::collections::HashSet;
+use symbolic::common::Name;
+use symbolic::demangle::{Demangle, DemangleOptions};
 
 pub struct RubySpy {
     process: Process,
@@ -14,15 +16,16 @@ pub struct RubySpy {
     ruby_vm_addr_location: usize,
     global_symbols_addr_location: Option<usize>,
     stack_trace_function: crate::core::types::StackTraceFn,
-    native_tracing: bool,
-    #[cfg(all(feature = "native_profiling", target_os = "linux", target_arch = "x86_64"))]
+    native_profiling: bool,
+    visited_ips: HashSet<u64>,
+    #[cfg(any(windows, target_os = "linux"))]
     unwinder: remoteprocess::Unwinder,
-    #[cfg(all(feature = "native_profiling", target_os = "linux", target_arch = "x86_64"))]
-    symbolicator: remoteprocess::Symbolicator
+    #[cfg(any(windows, target_os = "linux"))]
+    symbolicator: remoteprocess::Symbolicator,
 }
 
 impl RubySpy {
-    pub fn new(pid: Pid, force_version: Option<String>, native_tracing: bool) -> Result<Self> {
+    pub fn new(pid: Pid, force_version: Option<String>, native_profiling: bool) -> Result<Self> {
         #[cfg(all(windows, target_arch = "x86_64"))]
         if is_wow64_process(pid).context("check wow64 process")? {
             return Err(format_err!(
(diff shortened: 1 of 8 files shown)


src/core/ruby_spy.rs
```diff
@@ -123,6 +165,103 @@ impl RubySpy {
...
+        let is_cfunc = ruby_frame.name.contains("[c function]");
+        let mut found_start_of_cfunc = false || (i == 0 && is_cfunc);
+        if is_cfunc {
+            while let Some(native_frame) = native_frames_iter.next() {
+                if native_frame.name == "vm_call_cfunc_with_frame" {
+                    native_frames_iter.next();
+                    break;
+                }
+
+                // TODO: find a way to filter out internal ruby core functions, and make it configurable?
```
acj, line 251: Can you give an example of these?

src/core/ruby_spy.rs
```diff
@@ -123,6 +165,103 @@ impl RubySpy {
...
+fn merge_ruby_and_native_frames(
+    mut ruby_trace: StackTrace,
+    native_frames: Vec<StackFrame>,
+) -> StackTrace {
+    let ruby_frames = ruby_trace.trace;
+    let mut native_frames_iter = native_frames.iter();
+    let mut frames = vec![];
+    for (i, ruby_frame) in ruby_frames.iter().enumerate() {
+        let is_cfunc = ruby_frame.name.contains("[c function]");
+        let mut found_start_of_cfunc = false || (i == 0 && is_cfunc);
```
acj, line 243:
```suggestion
        let mut found_start_of_cfunc = i == 0 && is_cfunc;
```

src/core/ruby_spy.rs
```diff
@@ -123,6 +165,103 @@ impl RubySpy {
...
+        drop(lock);
+
+        let mut native_frames: Vec<StackFrame> = Vec::new();
+        for ip in native_ips {
+            self.symbolicate(ip, &mut |sf| {
+                let name: &str = sf
+                    .function
+                    .as_ref()
+                    .map(|s| s as &str)
+                    .unwrap_or_else(|| "cfunc (unnamed)".into());
```
acj, line 190: If there's a chance that this will be printed, it should probably use the "[c function]" notation like the existing C func code does.

src/core/ruby_spy.rs
```diff
@@ -123,6 +165,103 @@ impl RubySpy {
...
+
+    fn symbolicate(
+        &mut self,
+        ip: u64,
+        func: &mut dyn FnMut(&remoteprocess::StackFrame),
+    ) -> Result<(), remoteprocess::Error> {
+        match self.symbolicator.symbolicate(ip, true, func) {
+            Ok(_) => Ok(()),
+            Err(e) => {
+                // Do not reload more than once per IP which is not found
```
acj, line 221: Just curious, is this a performance optimization? Or does it cause an error if we reload more than once per IP?

src/main.rs
```diff
@@ -437,6 +446,8 @@ impl Args {
                 };
 
                 let no_drop_root = submatches.occurrences_of("no-drop-root") == 1;
+                let native_profiling = submatches.occurrences_of("native-profiling") == 1;
```
acj, line 449: Might be able to use `is_present` here like we do for the boolean options below.

acj (commented):
Thanks for this! The implementation looks good. It doesn't currently build on macOS because some functions and variables are hidden by the `cfg` directives.

I'm also thinking about test coverage. Maybe we can do something similar to the existing tests that load a core dump and get a stack trace from it. If you have ideas, feel free to share and/or try them.

Next steps:
- Please rebase onto main (should be easy - minor fixes and dependency upgrades), and then I'll approve the build so that CI can tell us what's broken. Let me know if you'd like help with the macOS build
- I'll put this through its paces in my Linux VM and will think about testing strategies a bit more

acj:
Closing for inactivity, but let me know if you're still interested in pursuing this.

PR #380, opened 2023-04-17
sec: Remove vulnerable time-0.1.x chrono dependency

ixti:
The dependency is optional for chrono and enabled by default for backward compatibility only.

See: https://rustsec.org/advisories/RUSTSEC-2020-0071
See: https://github.com/chronotope/chrono/blob/v0.4.23/CHANGELOG.md#0416

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ members = [
 
 [dependencies]
 anyhow = "1.0.58"
-chrono = "0.4.19"
+chrono = { version = "0.4.24", default-features = false, features = ["clock", "std"] }
 clap = "3.1.6"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -21,7 +21,7 @@ members = [
 
 [dependencies]
 anyhow = "1.0.58"
-chrono = "0.4.19"
+chrono = { version = "0.4.24", default-features = false, features = ["clock", "std", "wasmbind"] }
```
ixti, line 24: I believe we can omit `wasmbind` as well

acj (approved):
Thanks!

PR #381, opened 2023-05-28
Use Dependabot to maintain dependencies

p0mm:
## Would you like to use Dependabot? 
- Dependabot maintains dependencies: it does so by raising a PR to update to a new version
- github native, free and easy to use
- [more about Dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates)

In this PR I've configured Dependabot to check cargo and github-actions for updates weekly - this can be adapted very easily!

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


acj:
Thanks for the suggestion. We use Dependabot for some things already, but general dependency upgrades are handled as part of our release process.

PR #391, opened 2024-10-30
Remove unused direct dependency on goblin

musicinmybrain:
I was going to update goblin from 0.8 to 0.9, but it turns out it’s not used directly in this repository at all, so I removed the direct dependency instead.

Indirect dependencies exist via https://github.com/rbspy/rbspy-testdata and https://github.com/rbspy/spytools, but these were on 0.7.x rather than 0.8.x, so removing the direct dependency on 0.8.x at least keeps us from having to compile two versions of goblin. I still plan to follow up with PR’s to update both of those to 0.9.x, if possible.


acj:
Thanks!

PR #396, opened 2025-06-02
fix: do not stop sampling if Some(None) stacktrace is observed

korniltsev:


korniltsev:
I found this in pyroscope where rbspy is used as a library. There it manifests as just no stacktraces to the user of the library. There I start a puma server, it becomes idle (no requests), all samplers are exiting due to incorrect break, later when puma is not idle, no stacktraces is sent.

korniltsev:
Out of curiosity: is there a reason to use `continue` and skip the sleeping ?

acj:
No, that was my mistake. I spent some time trying to develop a test for this scenario and misremembered where the sleep code was. Thanks for asking 👍

PR #397, opened 2025-08-20
Include ruby class names in frame name output for ruby 3.3.0+

### LINKED ISSUE #379: Is it possible to include class / module names in output?
Compared to stackprof, generally I find rbspy profiles a bit harder for developers to read and interpret:

Here is stackprof for a particular stack trace in puma:

Screenshot 2023-03-23 at 11 51 09 AM

Same code path but rbspy:

Screenshot 2023-03-23 at 11 51 19 AM

Interpreting ruby callchains is much harder to do without understanding what class/module the methods belong to, as they tend to provide a lot of extra context.

In both cases we have the line numbers (interestingly, they don't match, but that's another thing to investigate) which can provide a lot of extra context, but nevertheless, the stackprof profiles are easier to action.

Is there any way that rbspy can deduce the class/module name? This is the function stackprof calls https://github.com/ruby/ruby/blob/50a709fb9e07f523c8a0649d3039161b77f300ad/vm_backtrace.c#L1578, and I **think** the code for this is in this guy https://github.com/ruby/ruby/blob/0112a5b3423ddb2ddf46c4392f23f9b75128568e/vm_insnhelper.c#L703 but I haven't looked too deeply. I'm curious if we can do something similar in rbspy and if not, why not?

dalehamel:
Fixes #379

Updates bindgen ruby structs for 3.3.0+ to include the necessary structs for parsing out the class name. 3.3.0 was the first release to include https://github.com/ruby/ruby/commit/abff5f62037284024aaf469fc46a6e8de98fa1e3, which makes looking up the classpath trivial compared to prior versions, where it was stored as an instance variable.

I used the method that @acj had been pursuing in his comment on #379, but updated it for ruby refactoring over the last 2 years. I copied the algorithm that ruby/ruby is doing, I was able to leverage some code that already existed for checking the method entry for c functions.

Note - I had to update the test fixtures / sample stacks for ruby 3.3.0+ as they now include the class name! (`Object#`, defined at the top level).

Here's a basic test program, and i've added a more complete one in https://github.com/rbspy/rbspy-testdata/pull/7, but I had to fix a bug in how we were reading the absolute path in order to get that to work.

```ruby
class Foo
  class Bar
    def bar()
      loop {}
    end
  end
end
Foo::Bar.new.bar
```

```
$ sudo cargo run -- snapshot --pid 75198
<main> - /Users/dalehamel/src/github.com/rbspy/rbspy/test.rb:15
Foo::Bar#bar - /Users/dalehamel/src/github.com/rbspy/rbspy/test.rb:12
loop - unknown:173
```


src/core/ruby_version.rs
```diff
@@ -919,6 +919,88 @@ macro_rules! get_stack_frame_2_5_0(
     )
 );
 
+macro_rules! get_stack_frame_3_3_0(
```
dalehamel, line 922: The classext approach used here is only valid in ruby 3.3.0+, it may be possible to get the info in another way in earlier versions, but the struct i'm using wasn't added until 3.3.0

src/core/ruby_version.rs
```diff
@@ -919,6 +919,88 @@ macro_rules! get_stack_frame_2_5_0(
...
+    () => (
+        fn get_stack_frame<T>(
+            iseq_struct: &rb_iseq_struct,
+            cfp: &rb_control_frame_t,
+            source: &T,
+        ) -> Result<StackFrame> where T: ProcessMemory {
+            let mut ep = cfp.ep as *mut usize;
```
dalehamel, line 929: Here to 935 is copy-pasta from above where it gets the cfunc names, because it did the exact algo i needed (copying ruby/ruby)

src/core/ruby_version.rs
```diff
@@ -919,6 +919,88 @@ macro_rules! get_stack_frame_2_5_0(
...
+            let rstring: RString = source.copy_struct(body.location.label as usize)
+                .context("couldn't copy RString")?;
+            let (path, absolute_path) = get_ruby_string_array(
+                body.location.pathobj as usize,
+                rstring.basic.klass as usize,
+                source
+            ).context("couldn't get ruby string from iseq body")?;
+
+            let method_name = get_ruby_string(body.location.label as usize, source)?;
+            let name = if class_name != "" {
```
dalehamel, line 982: again this is a bit lazy but it is simple for composing the class name with the method name.

I guess it needs to be a bit smarter to match https://github.com/ruby/ruby/blob/master/vm_backtrace.c#L1974-L1975, and check if it is a singleton or not and use "." instead of "#"? I'll see if i can add that easily enough, seems like it uses a bitmask on the flags attribute or something?

dalehamel:
FYI i think there is a legit issue here, but haven't pinned it down yet - the issue i'm seeing with the coredump tests seems to be related to reading the class name. We might need some more checks for safety, will try and repro more locally.

dalehamel:
> FYI i think there is a legit issue here, but haven't pinned it down yet - the issue i'm seeing with the coredump tests seems to be related to reading the class name. We might need some more checks for safety, will try and repro more locally.

Ok i think i've fixed this, I also refactored reading the class names into their own function, and improved the error handling.

src/core/ruby_version.rs
```diff
@@ -664,11 +664,95 @@ macro_rules! get_ruby_string_array_3_2_0(
...
             // pointer here
             let abs_path = get_ruby_string(abs_path_addr, source)
                 .unwrap_or(String::from("unknown"));
+
+            Ok((rel_path, abs_path))
+        }
+    )
+);
+
+macro_rules! get_ruby_string_array_3_3_0(
```
dalehamel, line 679: Needed to add these as I was getting a string allocation error when trying to read the absolute path in my coredump, rstring was reporting its size as enormous (might have been reported as max string length, or max string length -1?)

I added a sanity test where we can read the value only if it doesn't exceed a maximum size, and then pass 4096 as that is the maximum path size on linux.

src/core/ruby_version.rs
```diff
@@ -664,11 +664,95 @@ macro_rules! get_ruby_string_array_3_2_0(
...
+            // TODO: this assumes that the array contents are stored inline and not on the heap
+            // I think this will always be true but we should check instead
+            // the reason I am not checking is that I don't know how to check yet
+            let path_addr: usize = unsafe { rarray.as_.ary[0] as usize }; // 0 => relative path
+            let abs_path_addr: usize = unsafe { rarray.as_.ary[1] as usize }; // 1 => absolute path
+
+            let rel_path = get_ruby_string(path_addr, source)?;
+            // In the case of internal ruby functions (and maybe others), we may not get a valid
+            // pointer here
+            let abs_path = get_ruby_string_limit(abs_path_addr, 4096, source)
```
dalehamel, line 753: function is identical to 3_2 except for this line.

dalehamel:
freebsd failure seems like a flake

Cargo.toml
```diff
@@ -57,7 +57,7 @@ winapi = { version = "0.3.9", features = ["timeapi", "wow64apiset"] }
 
 [dev-dependencies]
 byteorder = "1.4.3"
-rbspy-testdata = "0.2.2"
+rbspy-testdata = { git = "https://github.com/dalehamel/rbspy-testdata.git", branch = "class-data" }
```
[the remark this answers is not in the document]
acj: I just tagged and published 0.2.4 with the enriched 3.3.0 coredump.

acj:
I think the FreeBSD failure is a workflow bug that I introduced recently. I'll take a closer look.

acj:
Okay, the CI bug should be fixed on main, so please rebase. I ran your branch [here](https://github.com/acj/rbspy/actions/runs/17351035121) with the testdata version bump and everything was green.

I need to review the version-specific code more closely and do some local testing, but I'm optimistic that we'll be able to merge this in the next week. Thanks for working on it!

dalehamel:
Thanks @acj

> Okay, the CI bug should be fixed on main, so please rebase. 

FYI looks like you only fixed that on your fork, didn't push to main on rbspy/rbspy yet. I fetched and rebased **your** main though.

> I ran your branch [here](https://github.com/acj/rbspy/actions/runs/17351035121) with the testdata version bump and everything was green.

Awesome, I moved the testdata back to the versioned crate. Commit looks the same as https://github.com/acj/rbspy/commit/9cb698348d44d6802a56b6bafaacee65b90c14a5, including the lockfile bumps.

> I need to review the version-specific code more closely and do some local testing, but I'm optimistic that we'll be able to merge this in the next week. Thanks for working on it!

Sounds great! Please give it a good look and if you can, try it on a real workload - I haven't had a chance to yet.

src/core/ruby_version.rs
```diff
@@ -919,6 +1003,107 @@ macro_rules! get_stack_frame_2_5_0(
...
+                    break;
+                }
+                unsafe {
+                    ep = env_specval.clone() as *mut usize;
+                    env_specval = source.copy_struct(ep.offset(-1) as usize).context(ep.offset(-1) as usize)?;
+                    env_me_cref = source.copy_struct(ep.offset(-2) as usize).context(ep.offset(-2) as usize)?;
+                }
+            }
+
+            let class_name = match source.copy_struct(env_me_cref).context(env_me_cref) {
```
acj, line 1043: This flattened version leans on the `?` operator and seems to work:

```rust
let imemo: rb_method_entry_struct = source.copy_struct(env_me_cref).context(env_me_cref)?;
let klass: RClass_and_rb_classext_t = source.copy_struct(imemo.defined_class).context(imemo.defined_class)?;
if klass.classext.classpath == 0 {
    return Err(anyhow::anyhow!("classpath was empty"));
}
let class_name = get_ruby_string(klass.classext.classpath as usize, source)?;
Ok(class_name.to_string())
```

acj (commented):
This has worked well in my local testing. I'll try to find a good production load to test it on this week, but that doesn't need to hold us up.

src/core/ruby_version.rs
dalehamel: @acj thanks i'll apply it and give it a go

acj (approved):

Issue #399, opened 2025-09-29
[bug] Mislabeled frames when class is detected


dalehamel:
In #397 i added detected of classpath, however, after using it more in production i've realized that the logic is a bit too simplified.

We can end up getting frames with method labels that don't make sense or don't match what ruby's own backtrace frame labeling would do.

As a general guiding principle, i think we should try and copy the logic in https://github.com/ruby/ruby/blob/v3_4_6/vm_backtrace.c#L1975.

I'm currently aware of a few deficiencies:

- If we have a label like `block in X`, we shouldn't join this with the class name or we get weird entries like `MyClass#block in X`
- We should detect that the class is a singleton class, we should use `.` as a joining character, not `#`
- We do not handle iclass types (this is relatively easy to support, we just need to redirect from klass to klass->klass before getting the classpath)
- We do not handle modules (again should be easy, the logic appears to be the same as for a class)
- We do not get the classpath for cfunc frames (this should get the classpath of the `owner` attribute, just need to refactor the code for looking up classpath and call it in this case)

I plan to put up a PR to fix these deficiencies, the first one is probably the most glaring.

I'd like to also create an improved test program that includes all these different edge cases. The test program i added was a bit naive and covered the basic cases of class of class owning a method, but we see a lot more complex stuff in the wild with ruby on rails.

dalehamel:
i think the root cause of this is that rbspy reads the `label` instead of the `base_label` for the method name to use.

This, detecting singletons, and walking to `klass->klass` on iclass's should be easy to fix. Hoping to get this done in the next week or so.

ivoanjo:
> I'd like to also create an improved test program that includes all these different edge cases. 

👋 for what it's worth https://github.com/ivoanjo/backtracie/blob/main/spec/unit/interesting_backtrace_helper.rb tries to be as evil as possible and exercise as many variants as we could think of for these kinds of things, do feel free to use it/steal ideas :)

dalehamel:
@ivoanjo oh man sweet, i'm going to use this for sure. May steal it for a test script with my work on the otel-bpf-profiler too :)

acj:
Fixed in [0.38.1](https://github.com/rbspy/rbspy/releases/tag/v0.38.1).

PR #400, opened 2025-10-04
Read all necessary attributes to properly get labels for CME, Iseq, and cfunc

dalehamel:
Draft fix for #399 

TODO

- table tests for qualified name and full label calculation
- Fix cfunc CME traversal and add classpath support to it
- Update coredump tests
- Create a more pathological case, possibly using https://github.com/ivoanjo/backtracie/blob/main/spec/unit/interesting_backtrace_helper.rb to cover additional cases

Verify the resulting method names match what is seen in the same workload for stackprof/vernier

### DIFF
--- a/src/core/ruby_version.rs
+++ b/src/core/ruby_version.rs
@@ -1004,22 +1004,41 @@ macro_rules! get_stack_frame_2_5_0(
 );
 
 macro_rules! get_stack_frame_3_3_0(
+
     () => (
         fn get_classname<T>(
             ep: &*const usize,
             source: &T,
-        ) -> Result<String> where T: ProcessMemory {
+        ) -> Result<(String, bool)> where T: ProcessMemory {
+            //https://github.com/ruby/ruby/blob/c149708018135595b2c19c5f74baf9475674f394/include/ruby/internal/value_type.h#L114¬
+            const RUBY_T_CLASS: usize = 0x2;
+            // https://github.com/ruby/ruby/blob/c149708018135595b2c19c5f74baf9475674f394/include/ruby/internal/value_type.h#L115C5-L115C74¬
+            const RUBY_T_MODULE: usize = 0x3;
+            //https://github.com/ruby/ruby/blob/c149708018135595b2c19c5f74baf9475674f394/include/ruby/internal/value_type.h#L138¬
+            const RUBY_T_ICLASS: usize = 0x1c;
+            // https://github.com/ruby/ruby/blob/c149708018135595b2c19c5f74baf9475674f394/include/ruby/internal/value_type.h#L142¬
+            const RUBY_T_MASK: usize = 0x1f;
+            // https://github.com/ruby/ruby/blob/1d1529629ce1550fad19c2d9410c4bf4995230d2/include/ruby/internal/fl_type.h#L158¬
+            const RUBY_FL_USHIFT: usize = 12;
+            // https://github.com/ruby/ruby/blob/1d1529629ce1550fad19c2d9410c4bf4995230d2/include/ruby/internal/fl_type.h#L323-L324¬
+            const RUBY_FL_USER1: usize = 1 << (RUBY_FL_USHIFT + 1);
+            // https://github.com/ruby/ruby/blob/1d1529629ce1550fad19c2d9410c4bf4995230d2/include/ruby/internal/fl_type.h#L394¬
+            const RUBY_FL_SINGLETON: usize = RUBY_FL_USER1;
+
             let mut ep = ep.clone() as *mut usize;
+            let mut singleton = false;
 
-            let frame_flag: usize = unsafe {
-                source.copy_struct(ep.offset(0) as usize).context(ep.offset(0) as usize)?
-            };
+            //let frame_flag: usize = unsafe {
+            //    source.copy_struct(ep.offset(0) as usize).context(ep.offset(0) as usize)?
+            //};
 
-            // VM_FRAME_MAGIC_MASK   = 0x7fff0001
-            // VM_FRAME_MAGIC_METHOD = 0x11110001
-            if (frame_flag & 0x7fff0001) != 0x11110001 {
-                return Err(anyhow::anyhow!(format!("frame flag {:x} does not match method mask, cannot read class", frame_flag)));
-            }
+
+            // TODO verify if this is needed, i don't think it is
+            //// VM_FRAME_MAGIC_MASK   = 0x7fff0001
+            //// VM_FRAME_MAGIC_METHOD = 0x11110001
+            //if (frame_flag & 0x7fff0001) != 0x11110001 {
+            //    return Err(anyhow::anyhow!(format!("frame flag {:x} does not match method mask, cannot read class", frame_flag)));
+            //}
 
             let mut env_specval: usize = unsafe {
                 source.copy_struct(ep.offset(-1) as usize).context(ep.offset(-1) as usize)?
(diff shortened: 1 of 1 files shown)


ci/ruby-programs/cme_complex_labels.rb
```diff
@@ -0,0 +1,26 @@
+require_relative './interesting_backtrace_helper'
```
dalehamel, line 1: Some preamble to test different cases, then load the test file from backtracie

.github/workflows/coredump.yml
```diff
@@ -40,16 +40,20 @@ jobs:
...
       - name: Inspect process
         run: sudo ./target/release/rbspy inspect --pid ${{ steps.ruby-script.outputs.pid }}
 
+      # default coredump filter is 0x33, update it to basically dump all memory here
+      # to best mimic the real process otherwise heap data like strings may not
+      # be accessible
       - name: Create core dump
         shell: sudo bash {0}
         run: |
+          echo 0x7f > /proc/${{ steps.ruby-script.outputs.pid }}/coredump_filter
```
dalehamel, line 56: Note that 0xff also seems to work, and it might also be a good idea to do gcore -a 

It drove me crazy trying to figure out why my string reads would work on a local process, but fail in the coredump - the memory region the string was in was not in the coredump lol.

Github uses 0x33 by default for coredump filter.

src/core/ruby_version.rs
```diff
@@ -416,7 +425,29 @@ macro_rules! get_stack_trace(
             let mut trace = Vec::new();
             let cfps = get_cfps(thread.cfp as usize, stack_base(&thread) as usize, source)?;
             for cfp in cfps.iter() {
-                if cfp.iseq as usize == 0 {
+                let cfunc = ruby_frame_cfunc(&cfp, source)?;
```
dalehamel, line 428: I flipped the conditional around here and got rid of the frame flag filtering to match what ruby's own backtrace function is doing.

src/core/ruby_version.rs
```diff
@@ -485,13 +493,13 @@ macro_rules! get_stack_trace(
         fn stack_base(thread: &$thread_type) -> i64 {
             // Ruby stack grows down, starting at
             //   ruby_current_thread->stack + ruby_current_thread->stack_size - 1 * sizeof(rb_control_frame_t)
-            // I don't know what the -1 is about. Also note that the stack_size is *not* in bytes! stack is a
-            // VALUE*, and so stack_size is in units of sizeof(VALUE).
+            // We must skip two dummy frames:
+            // https://github.com/ruby/ruby/blob/5445e0435260b449decf2ac16f9d09bae3cafe72/vm_backtrace.c#L477-L485%C2%A
```
dalehamel, line 497: Note that i think the other -1 is occurring implicitly in our calculation of how many CFPs to collect, hence why this is -1 and not -2 here.

src/core/ruby_version.rs
```diff
@@ -1003,81 +1011,293 @@ macro_rules! get_stack_frame_2_5_0(
...
-                return Err(anyhow::anyhow!("classpath was empty"));
+            if klass == 0 {
+                return Err(anyhow::anyhow!("klass was empty"));
+            }
+
+            let class_path = rb_class_path(klass, source)?;
+            Ok((class_path, singleton))
+        }
+
+        // TODO make some tests for profile_full_label_name to cover the various cases it needs
```
dalehamel, line 1201: TODO - delete these todos

src/core/ruby_version.rs
```diff
@@ -1003,81 +1011,293 @@ macro_rules! get_stack_frame_2_5_0(
...
+                source.copy_struct(iseq.body as usize)
+                    .context("couldn't copy rb_iseq_constant_body")?;
+            let frame_flag: usize = unsafe {
+                source.copy_struct(cfp.ep.offset(0) as usize).context(cfp.ep.offset(0) as usize)?
+            };
             let rstring: RString = source.copy_struct(body.location.label as usize)
                 .context("couldn't copy RString")?;
+
+            let mut method_name = "".to_string();
+            if body.local_iseq != std::ptr::null_mut() {
```
dalehamel, line 1279: We should be able to depend on this field being there most of the time, but to be safe we add this defensive check.

TODO add documentation link for why we do the indirection to local_iseq here.

src/core/ruby_version.rs
```diff
@@ -1289,71 +1516,126 @@ macro_rules! get_cfunc_name_unsupported(
...
             };
 
-            // #define VM_ENV_FLAG_LOCAL 0x02
-            while env_specval & 0x02 != 0 {
-                if !check_method_entry(env_me_cref, source)?.is_null() {
-                    break;
+            if env_flags & VM_ENV_FLAG_LOCAL == 0 {
+                //println!("ALREADY LOCAL {:X}", env_me_cref);
+            }
+            while env_flags & VM_ENV_FLAG_LOCAL != 0 {
```
dalehamel, line 1563: Note the conditional should have been checking the flags, not the specval.

src/core/ruby_version.rs
```diff
@@ -1289,71 +1516,126 @@ macro_rules! get_cfunc_name_unsupported(
...
                 }
             }
 
-            let imemo: rb_method_entry_struct = source.copy_struct(env_me_cref).context(env_me_cref)?;
+            let me = check_method_entry(env_me_cref, true, source)?;
+            //println!("ME {:X} {:X}", me, env_me_cref);
+            Ok(me)
+        }
+
+        fn check_method_entry<T: ProcessMemory>(
```
dalehamel, line 1585: This function was also pretty broken. I rewrote it to mimic the logic in ruby's vm_insnhelper.c

src/core/ruby_version.rs
```diff
@@ -1289,71 +1516,126 @@ macro_rules! get_cfunc_name_unsupported(
...
+            let (class_path, singleton) = get_classpath(cme, true, source).unwrap_or(("".to_string(), false));
+
+            let imemo: rb_method_entry_struct = source.copy_struct(cme).context(cme)?;
             if imemo.def.is_null() {
                 return Err(format_err!("No method definition").into());
             }
 
-            let ttype = ((imemo.flags >> 12) & 0x07) as usize;
+
+            // FIXME - i'm pretty sure this mask is wrong, it should be 0xff
```
dalehamel, line 1632: TODO - delete me, i fixed the mask.

src/core/ruby_version.rs
```diff
@@ -1749,56 +2030,64 @@ mod tests {
...
+                lineno: Some(21),
             },
             StackFrame {
-                name: "loop".to_string(),
+                name: "Kernel#loop".to_string(),
                 relative_path: "<internal:kernel>".to_string(),
                 absolute_path: Some("unknown".to_string()),
                 lineno: Some(192),
             },
+            StackFrame {
```
dalehamel, line 2076: The coredump was invalid and this frame was dropped entirely. I redid the coredump and added the missing frame.

src/core/ruby_version.rs
```diff
@@ -2658,8 +3000,8 @@ mod tests {
     #[test]
     fn test_get_ruby_stack_trace_3_3_1() {
         let source = coredump_3_3_0();
-        let vm_addr = 0x7f7ff21f1868;
```
dalehamel, line 2661: These and others needed to be updated since I replaced the coredump with a non-truncated on.

src/core/ruby_version.rs
```diff
@@ -2925,4 +3288,64 @@ mod tests {
         .unwrap();
         assert_eq!(real_stack_trace_3_3_0(), stack_trace.unwrap().trace);
     }
+
+    #[rstest]
```
dalehamel, line 3292: I added rstest here for table tests on these label calculation functions.

The logic of the functions themselves **should** match their counterparts in cruby. 

In case we find cases where they don't, we can fix it and add regression tests here based on the input combos.

There is no need to update the test, just add new cases to the macro with the inputs to use and expected output.

src/core/ruby_version.rs
```diff
@@ -1003,81 +1011,291 @@ macro_rules! get_stack_frame_2_5_0(
...
             source: &T,
         ) -> Result<StackFrame> where T: ProcessMemory {
+            let cme = locate_method_entry(&cfp.ep, source)?;
+            let mut class_path = "".to_string();
+            let mut singleton = false;
+            let iseq = if cme != 0 {
+                let imemo: rb_method_entry_struct = source.copy_struct(cme).context(cme)?;
+                let method_type = source.copy(imemo.def as usize, 1).context(imemo.def as usize)?;
+                // https://github.com/ruby/ruby/blob/v3_3_0/internal/imemo.h#L21
+                let method_type = method_type[0] & 0xf;
```
dalehamel, line 1247: Note 0xf is because it is a bitfield of 4 bits, so we read 1 byte and then mask out the upper bits.

dalehamel:
> This looks great. Really appreciate the effort to fix and clean up the various helpers, and also the detailed comments along the way. Go ahead and clean up the commits when you have time, and I'll do some local testing to hopefully shake loose any edge cases. Then I'm good with merging.

Hey @acj hope you are well, i'm back from vacation and i've addressed your style nits and squashed a few commits. LMK if you found anything that needs to be addressed.

acj (approved):

